PR #10, opened 2023-07-13
XIVY-11751 provide standalone integration

ivy-lmu:
she is not yet a beauty, but she works
image

### DIFF
--- a/packages/editor/src/yaml-table.tsx
+++ b/packages/editor/src/yaml-table.tsx
@@ -13,13 +13,24 @@ import './codicon.css';
 import './yaml-table.css';
 
 interface YAMLVariablesTableProps {
-    vscodeApi: WebviewApi<unknown>;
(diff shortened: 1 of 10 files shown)


packages/editor/src/yaml-table.tsx
```diff
@@ -13,13 +13,24 @@ import './codicon.css';
 import './yaml-table.css';
 
 interface YAMLVariablesTableProps {
-    vscodeApi: WebviewApi<unknown>;
+    vscodeApi?: WebviewApi<unknown>;
```
ivy-lli, line 16: I assume this part needs to be part to from the vscode integration?

ivy-lli (approved):
🆒

packages/editor/src/yaml-table.tsx
ivy-lmu: Yes, this API is provided when the table is rendered in the vs code extension.

PR #39, opened 2024-03-27
XIVY-11980 implement basic editor layout

ivy-lgi:


packages/editor/src/components/editor/Editor.tsx
```diff
@@ -0,0 +1,39 @@
...
+    <Flex justifyContent='center' alignItems='center' style={{ height: '100%' }}>
+      <span>Content</span>
+    </Flex>
+  );
+  const detailTitle = 'Variables Editor - variableA';
+  const detailContent = (
+    <Flex justifyContent='center' alignItems='center' style={{ height: '100%' }}>
+      <span>Properties</span>
+    </Flex>
+  );
```
ivy-lli, line 18: better directly craft some components, then you can already use them in the jsx below.

ivy-lli (approved):
Fine for me, but if you merge it like this, you will break the integration in the vscode extension. 
You could also leave the yaml-table component as it is for now and simply integrate it into the content part of the editor, or maybe just disable the vscode integration tests for now.

ivy-lgi:
I moved the new editor into its own package to not interfere with existing stuff.

PR #40, opened 2024-03-27
XIVY-11980 implement Variables component

ivy-lgi:
image

### DIFF
--- a/packages/new-editor/src/components/editor/Editor.css
+++ b/packages/new-editor/src/components/editor/Editor.css
@@ -1,7 +1,3 @@
-.editor {
-  height: 100vh !important;
-}
-
 .master-panel {
   background-color: var(--N100);
 }
--- a/packages/new-editor/src/components/editor/Editor.css
+++ b/packages/new-editor/src/components/editor/Editor.css
@@ -1,3 +1,7 @@
+.editor {
+  height: 100vh !important;
+}
+
 .master-panel {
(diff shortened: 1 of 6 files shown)


packages/new-editor/src/components/editor/Editor.css
```diff
@@ -1,3 +1,7 @@
+.editor {
+  height: 100vh !important;
```
ivy-lli, line 2: sorry but I don't like to see `!important` statements in CSS...
IMO this is mostly not needed and only a sign for missing knowledge how CSS and the specificity of selectors work: https://1ivy.atlassian.net/wiki/spaces/IVYTEAM/pages/6988639/CSS

packages/new-editor/src/components/editor/Editor.tsx
```diff
@@ -1,35 +1,44 @@
...
   const masterTitle = 'Rest Clients Editor';
-  const masterContent = (
-    <Flex justifyContent='center' alignItems='center' style={{ height: '100%' }}>
-      <span>Content</span>
-    </Flex>
-  );
   const detailTitle = 'Variables Editor - variableA';
+  const variables = [
+    {
+      key: 'param.procurementRequest',
```
ivy-lli, line 12: as example I would rather use a variables.yaml structure than a mapping example ;)

ivy-lli (approved):

PR #41, opened 2024-03-28
XIVY-11980 implement VariableDetail component

ivy-lgi:
image

### DIFF
--- a/packages/new-editor/src/components/editor/variables/detail/Variable.tsx
+++ b/packages/new-editor/src/components/editor/variables/detail/Variable.tsx
@@ -1,19 +1,12 @@
 import { Fieldset, Input, SimpleSelect, Textarea, ToolbarTitle } from '@axonivy/ui-components';
-import { Metadata, type Variable } from '../../data/Variable';
+import { metadataOptions, type Variable } from '../../data/Variable';
 
 type VariableProps = {
   variable: Variable;
 };
 
 export const VariableDetail = ({ variable }: VariableProps) => {
-  const metadataOptions = Object.keys(Metadata)
-    .filter(key => isNaN(Number(key)))
-    .map(key => {
-      const option = key.toString();
-      return { value: option, label: option };
-    });
-  metadataOptions.unshift({ value: 'None', label: 'None' });
-  const selectedMetadataOption = variable.metadata === undefined ? 'None' : Metadata[variable.metadata];
+  const selectedMetadataOption = metadataOptions.find(option => option.value === variable.metadata);
 
   return (
     <>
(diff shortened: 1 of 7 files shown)


packages/new-editor/src/components/editor/variables/detail/Variable.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const VariableDetail = ({ variable }: VariableProps) => {
+  const metadataOptions = Object.keys(Metadata)
+    .filter(key => isNaN(Number(key)))
+    .map(key => {
+      const option = key.toString();
+      return { value: option, label: option };
+    });
+  metadataOptions.unshift({ value: 'None', label: 'None' });
+  const selectedMetadataOption = variable.metadata === undefined ? 'None' : Metadata[variable.metadata];
```
ivy-lgi, line 16: I don't like this enum handling stuff and having to introduce an extra option to not select anything. Any ideas on improvements?

packages/new-editor/src/components/editor/variables/detail/Variable.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  return (
+    <>
+      <ToolbarTitle>Variable Configuration</ToolbarTitle>
```
ivy-lli, line 20: why not sidebar header?

packages/new-editor/src/components/editor/variables/detail/Variable.tsx
ivy-lli: yes enums technially do not exist in ts. or better they really do not exist in js. ts does some "magic" here to fake them in js...

I would suppose something like in the inscription-view:
```ts
export const RESPONSIBLE_TYPE = {
  ROLE: 'Role',
  ROLE_FROM_ATTRIBUTE: 'Role from Attr.',
  USER_FROM_ATTRIBUTE: 'User from Attr.',
  DELETE_TASK: 'Nobody & delete'
} as const satisfies Record<WfActivatorType, string>;

Object.entries(RESPONSIBLE_TYPE)
        .filter(([value]) => !(props.optionFilter && props.optionFilter.includes(value as WfActivatorType)))
        .map(([value, label]) => ({ label, value }));
```

or in the form-editor:
```ts
const typeOptions: FieldOption<InputType>[] = [
  { label: 'Email', value: 'EMAIL' },
  { label: 'Password', value: 'PASSWORD' },
  { label: 'Text', value: 'TEXT' }
] as const;
```

at the moment you only "fake" the metas, so totally fine to hold this type as simple as possible.

ivy-lli (approved):

packages/new-editor/src/components/editor/variables/detail/Variable.tsx
ivy-lgi: This is the title underneath the header, not the header itself:
image

Although, maybe I should not use the `ToolbarTitle` component for this?

ivy-lli (approved):

PR #43, opened 2024-04-02
XIVY-11980 implement variables control

ivy-lgi:
- Buttons for adding and deleting variables are present but not functional
- Always show detail view 

image

### DIFF
--- a/packages/new-editor/src/components/editor/Editor.css
+++ b/packages/new-editor/src/components/editor/Editor.css
@@ -1,9 +1,9 @@
 .master-panel {
-  background-color: var(--N100);
+  background-color: var(--N75);
 }
 
 .master-toolbar {
-  background-color: var(--N100);
+  background-color: var(--N75);
   border-bottom: none;
 }
 
--- a/packages/new-editor/src/components/editor/Editor.css
+++ b/packages/new-editor/src/components/editor/Editor.css
@@ -2,11 +2,6 @@
   background-color: var(--N75);
 }
 
-.master-toolbar {
-  background-color: var(--N75);
-  border-bottom: none;
-}
-
 .content {
   margin: var(--size-3);
 }
--- a/packages/new-editor/src/components/editor/control/Control.tsx
+++ b/packages/new-editor/src/components/editor/control/Control.tsx
@@ -1,0 +1,21 @@
+import type { ReactElement } from 'react';
+import { Fragment, cloneElement } from 'react';
+import './Control.css';
+
+type ControlProps = {
+  buttons: ReactElement[];
+};
+
+export const Control = ({ buttons }: ControlProps) => {
+  return buttons.map((button, index) => {
+    if (index === 0) {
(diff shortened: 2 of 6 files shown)


packages/new-editor/src/components/editor/variables/detail/Variable.tsx
```diff
@@ -2,15 +2,20 @@ import { Fieldset, Input, SimpleSelect, Textarea, ToolbarTitle } from '@axonivy/
 import { metadataOptions, type Variable } from '../../data/Variable';
 
 type VariableProps = {
-  variable: Variable;
+  variable: Variable | undefined;
```
ivy-lli, line 5: instead of `| undefined` you can also just add a ? do the variable: `variable?: Variable`

packages/new-editor/src/components/editor/variables/detail/Variable.tsx
```diff
@@ -2,15 +2,20 @@ import { Fieldset, Input, SimpleSelect, Textarea, ToolbarTitle } from '@axonivy/
...
 };
 
 export const VariableDetail = ({ variable }: VariableProps) => {
+  const title = <ToolbarTitle>Variable Configuration</ToolbarTitle>;
```
ivy-lli, line 9: I think you shouldn't reuse this toolbartitle here. it may causes unwanted changes, if we ever change this component and don't have this usage here in mind...

ivy-lli: instead you could just e.g. leave a message about select a variable or something like that.

packages/new-editor/src/components/editor/control/Control.tsx
```diff
@@ -0,0 +1,21 @@
...
+};
+
+export const Control = ({ buttons }: ControlProps) => {
+  return buttons.map((button, index) => {
+    if (index === 0) {
+      return cloneElement(button, { style: { marginLeft: 'auto' } });
+    }
+    return (
+      <Fragment key={index}>
+        <div className='vertical-line' />
```
ivy-lli, line 16: there is a seperator component in the ui-components. did you try to use it?

packages/new-editor/src/components/editor/Editor.css
```diff
@@ -1,9 +1,9 @@
 .master-panel {
-  background-color: var(--N100);
+  background-color: var(--N75);
 }
 
 .master-toolbar {
-  background-color: var(--N100);
+  background-color: var(--N75);
```
ivy-lli, line 6: I'm not sure how the figma mockup looks like, but IMO this toolbar should look the same as in the other editors even if we don't have all the buttons on it... 
More like the example here (also with the collapse sidebar function): 
https://jenkins.ivyteam.io/job/ui-components/job/master/106/artifact/packages/components/storybook-static/index.html?path=/story/master-toolbar--with-title

ivy-lli (approved):

ivy-lgi:
edit-variables

PR #44, opened 2024-04-04
XIVY-11980 implement variable delete control

ivy-lgi:
- The delete button is disabled if no variable is selected.
- After deleting a variable, the next variable is automatically selected.
  - If the deleted variable was the last variable in the list of children of its parent, the previous variable is selected instead.
  - If the deleted variable was the last remaining child of its parent altogether, the parent is selected instead.

implement-variable-delete-control

PR #46, opened 2024-04-04
XIVY-11980 implement variable add control

ivy-lgi:
Continuation of https://github.com/axonivy/config-editor-client/pull/44

implement-variable-add-control


weissreto (approved):
👍as far as I understand it

PR #47, opened 2024-04-04
XIVY-11980 minor changes

ivy-lgi:
- rename indexes to path
- use Array instead of []


ivy-lli (commented):
😅 👍 
Would be totally fine, if you change the Array stuff when you detect an appearance. Technically it makes no big deal, and we have many places where still the [] syntax is used so... But cool thanks :)

ivy-lli (approved):

PR #49, opened 2024-04-04
XIVY-11980 allow editing variables

ivy-lgi:
allow-variable-edit

### DIFF
--- a/packages/new-editor/src/components/editor/variables/detail/Variable.tsx
+++ b/packages/new-editor/src/components/editor/variables/detail/Variable.tsx
@@ -1,11 +1,16 @@
 import { Fieldset, Input, Message, SimpleSelect, Textarea } from '@axonivy/ui-components';
+import { getVariable } from '../../../../data/data';
+import type { TreePath } from '../../../../types/config';
 import { metadataOptions, type Variable } from '../../data/Variable';
 
 type VariableProps = {
-  variable?: Variable;
+  variables: Array<Variable>;
+  variablePath: TreePath;
+  setVariables: (variables: Array<Variable>) => void;
 };
 
-export const VariableDetail = ({ variable }: VariableProps) => {
+export const VariableDetail = ({ variables, variablePath, setVariables }: VariableProps) => {
+  const variable = getVariable(variables, variablePath);
   if (!variable) {
     return <Message>Select a variable to edit.</Message>;
   }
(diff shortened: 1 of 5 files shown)


packages/new-editor/src/components/editor/variables/detail/Variable.tsx
```diff
@@ -1,34 +1,49 @@
...
+    const newVariables = structuredClone(variables);
+    getVariable(newVariables, variablePath)![key] = value;
+    setVariables(newVariables);
+  };
+
   return (
     <>
       <Fieldset label='Name'>
-        <Input value={variable.name} />
+        <Input value={variable.name} onChange={event => handleVariableAttributeChange('name', event.target.value)} />
```
ivy-lgi, line 30: Refactoring the attribute name in the interface `Variable` will not change the string that is used to identify which attribute to change. TypeScript will however still show an error afterwards. Any way to make this work without having to create a function for each attribute?

packages/new-editor/src/components/editor/variables/detail/Variable.tsx
```diff
@@ -1,34 +1,49 @@
...
   if (!variable) {
     return <Message>Select a variable to edit.</Message>;
   }
 
   const selectedMetadataOption = metadataOptions.find(option => option.value === variable.metadata);
   const doesNotHaveChildren = variable.children.length == 0;
 
+  const handleVariableAttributeChange = (key: keyof Variable, value: any) => {
+    const newVariables = structuredClone(variables);
+    getVariable(newVariables, variablePath)![key] = value;
```
ivy-lli, line 23: better check the variable is available instead of just do `!`.

packages/new-editor/src/components/editor/variables/detail/Variable.tsx
```diff
@@ -1,34 +1,49 @@
...
+export const VariableDetail = ({ variables, variablePath, setVariables }: VariableProps) => {
+  const variable = getVariable(variables, variablePath);
   if (!variable) {
     return <Message>Select a variable to edit.</Message>;
   }
 
   const selectedMetadataOption = metadataOptions.find(option => option.value === variable.metadata);
   const doesNotHaveChildren = variable.children.length == 0;
 
+  const handleVariableAttributeChange = (key: keyof Variable, value: any) => {
```
ivy-lli, line 21: make it generic instead of just do any:
`const handleVariableAttributeChange = <TKey extends keyof Variable>(key: TKey, value: Variable[TKey]) => {`

https://www.typescriptlang.org/play?ssl=6&ssc=109&pln=6&pc=3#code/ATAuE8AcFNgNQIYCcCWCBGAbWBeYBvAKBBADsEBbaALmAGdRVSBzAbmJIBNo6BjVSKBQB7UrQZM2HAL4cOvUQ2AALBKU7ZEqDNgCCoRinQBXUNADCqlrmAAeACoBpaOGDQAHmfV1gAaxfCAGbwyGhY0AB8ABT+4LROLgA0wABuCJjGNCHa4QDaCeAAugCUwDgRBLIcVhrQWmF6BqgmZpZqzNBRAETkVF3JAIwATADMxaxAA

packages/new-editor/src/components/editor/variables/detail/Variable.tsx
ivy-lli: Not sure what you mean here

ivy-lli (approved):

weissreto (approved):
🔎👍

packages/new-editor/src/components/editor/variables/detail/Variable.tsx
ivy-lgi: The attribute to be updated is identified by a string (the name of the attribute). If I refactor the name of the attribute, this string does not change. I would like this string to update as well with the refactor.

refactoring-issue

PR #53, opened 2024-04-12
XIVY-11980 adjust variable detail to metadata

ivy-lgi:
metadataoptions

### DIFF
--- a/packages/new-editor/src/components/editor/variables/detail/EnumValues.tsx
+++ b/packages/new-editor/src/components/editor/variables/detail/EnumValues.tsx
@@ -1,19 +1,19 @@
 import { Button, Fieldset, InputCell, SelectRow, Table, TableBody, TableCell, useTableSelect } from '@axonivy/ui-components';
 import { IvyIcons } from '@axonivy/ui-icons';
 import { flexRender, getCoreRowModel, useReactTable, type ColumnDef } from '@tanstack/react-table';
-import { addRow, deleteFirstSelectedRow } from '../../../../utils/table';
 import { updateRowData } from '../../../../data/table-data';
 import { treeNodeValueAttribute } from '../../../../types/config';
+import { addRow, deleteFirstSelectedRow } from '../../../../utils/table';
 import { Control } from '../../control/Control';
 import { toEnumMetadataUpdate, type VariableUpdates } from '../../data/Variable';
 
-type PossibleEnumValuesProps = {
+type EnumValuesProps = {
   selectedValue: string;
   values: Array<string>;
   onChange: (updates: VariableUpdates) => void;
 };
 
-export const PossibleEnumValues = ({ selectedValue: value, values, onChange }: PossibleEnumValuesProps) => {
+export const EnumValues = ({ selectedValue: value, values, onChange }: EnumValuesProps) => {
   const selection = useTableSelect<string>();
   const columns: Array<ColumnDef<string, string>> = [
     {
@@ -24,7 +24,11 @@ export const PossibleEnumValues = ({ selectedValue: value, values, onChange }: P
   ];
   const meta = {
     updateData: (rowId: string, _columnId: string, value: unknown) => {
-      const newValues = updateRowData(values, Number(rowId), String(value));
+      const valueString = String(value);
+      if (values.includes(valueString)) {
+        return;
+      }
+      const newValues = updateRowData(values, Number(rowId), valueString);
       onChange([toEnumMetadataUpdate(newValues)]);
     }
   };
@@ -39,12 +43,12 @@ export const PossibleEnumValues = ({ selectedValue: value, values, onChange }: P
     meta: meta
   });
 
-  const addPossibleValue = () => {
+  const addValue = () => {
     const newValues = addRow(table, values, '');
     onChange([toEnumMetadataUpdate(newValues)]);
   };
 
-  const deletePossibleValue = () => {
+  const deleteValue = () => {
     const newValues = deleteFirstSelectedRow(table, values);
     const updates: VariableUpdates = [toEnumMetadataUpdate(newValues)];
     if (!newValues.includes(value)) {
@@ -55,12 +59,12 @@ export const PossibleEnumValues = ({ selectedValue: value, values, onChange }: P
 
   return (
     <Fieldset
-      label='List of possible Values'
+      label='List of possible values'
       control={
         <Control
           buttons={[
-            <Button key='addButton' icon={IvyIcons.Plus} onClick={addPossibleValue} />,
-            <Button key='deleteButton' icon={IvyIcons.Trash} onClick={deletePossibleValue} disabled={!table.getIsSomeRowsSelected()} />
+            <Button key='addButton' icon={IvyIcons.Plus} onClick={addValue} />,
+            <Button key='deleteButton' icon={IvyIcons.Trash} onClick={deleteValue} disabled={!table.getIsSomeRowsSelected()} />
           ]}
         />
       }
--- a/packages/new-editor/src/components/editor/variables/detail/Metadata.tsx
+++ b/packages/new-editor/src/components/editor/variables/detail/Metadata.tsx
@@ -1,17 +1,14 @@
 import { Fieldset, SimpleSelect } from '@axonivy/ui-components';
 import { treeNodeValueAttribute } from '../../../../types/config';
 import {
-  asEnumMetadata,
+  isEnumMetadata,
   metadataOptions,
-  metadataTypeDaytime,
-  metadataTypeEnum,
   variableMetadataAttribute,
-  type EnumMetadata,
   type MetadataType,
   type Variable,
   type VariableUpdates
 } from '../../data/Variable';
-import { PossibleEnumValues } from './PossibleEnumValues';
+import { EnumValues } from './EnumValues';
 
 type MetadataFieldsetProps = {
   variable: Variable;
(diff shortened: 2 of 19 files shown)


packages/new-editor/src/components/editor/variables/detail/Metadata.tsx
```diff
@@ -0,0 +1,51 @@
...
+      <Fieldset label='Metadata'>
+        <SimpleSelect
+          value={variable.metadata.type}
+          items={metadataOptions}
+          emptyItem={true}
+          onValueChange={(value: MetadataType) => {
+            const newMetadata = { type: value };
+            const updates: VariableUpdates = [];
+            switch (value) {
+              case metadataTypeDaytime:
```
ivy-lli, line 35: I think I know why you did hide the value behind a variable. It is because then you can quickly rename the values right?

Understandable, but IMO this makes the readability of the code much worse. 
If I compare `case metadataTypeDaytime` to `case 'daytime'` I can ready the second one much quicker. 
And if the reason above is the only one, I'm not sure if it is enough, because we cannot simply change those values. They are in variable.yaml files in the project so, we need to convert customers projects if we want to change these... 
So IMO I would value the code readability much higher here than the possibility of refactoring the code a little bit faster. 
Or do you have another reason why you do not simply use the values?

packages/new-editor/src/components/editor/variables/detail/Metadata.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const enumMetadata = asEnumMetadata(variable.metadata);
+
+  return (
+    <>
+      <Fieldset label='Metadata'>
+        <SimpleSelect
+          value={variable.metadata.type}
+          items={metadataOptions}
+          emptyItem={true}
+          onValueChange={(value: MetadataType) => {
```
ivy-lli, line 31: I would move this value change function outside of the JSX part.

packages/new-editor/src/components/editor/variables/detail/PossibleEnumValues.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { Control } from '../../control/Control';
+import { toEnumMetadataUpdate, type VariableUpdates } from '../../data/Variable';
+
+type PossibleEnumValuesProps = {
+  selectedValue: string;
+  values: Array<string>;
+  onChange: (updates: VariableUpdates) => void;
+};
+
+export const PossibleEnumValues = ({ selectedValue: value, values, onChange }: PossibleEnumValuesProps) => {
```
ivy-lli, line 16: I would call it simply `EnumValues`

packages/new-editor/src/components/editor/variables/detail/Variable.tsx
ivy-lli, line 1: I would also rename the file to 'VariablesDetail'

packages/new-editor/src/components/editor/variables/detail/Variable.tsx
```diff
@@ -1,50 +1,45 @@
...
   }
 
   const hasChildren = variableHasChildren(variable);
 
-  const handleVariableAttributeChange = <TKey extends keyof Variable>(key: TKey, value: Variable[TKey]) => {
-    const newVariables = structuredClone(variables);
-    updateNode(newVariables, variablePath, key, value);
+  const handleVariableAttributeChange = (updates: VariableUpdates) => {
+    const newVariables = updateNode(variables, variablePath, updates);
     setVariables(newVariables);
```
ivy-lli, line 24: be careful that you don't accidentally modify the state inside of updateNode! https://react.dev/reference/rules/components-and-hooks-must-be-pure#state

better use setVariable(old => updateNode(old, varialbePath, updates));

packages/new-editor/src/components/editor/variables/master/Variables.tsx
```diff
@@ -54,31 +54,29 @@ export const Variables = ({ variables, setVariables, setSelectedVariablePath }:
...
   };
 
   const deleteVariable = () => {
-    const newVariables = structuredClone(variables);
-    const selectedVariablePath = deleteFirstSelectedRow(table, newVariables);
-    setSelectedVariablePath(selectedVariablePath);
-    setVariables(newVariables);
+    const deleteFirstSelectedRowReturnValue = deleteFirstSelectedRow(table, variables);
+    setSelectedVariablePath(deleteFirstSelectedRowReturnValue.selectedVariablePath);
+    setVariables(deleteFirstSelectedRowReturnValue.newData);
```
ivy-lli, line 79: same here as above -> rule of react, do not modify state directly. 
structured clone would "saved" you from this. if you don't want to use it, use the lambda method from setVariable.

packages/new-editor/src/components/editor/variables/detail/PossibleEnumValues.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const updates: VariableUpdates = [toEnumMetadataUpdate(newValues)];
+    if (!newValues.includes(value)) {
+      updates.push({ key: treeNodeValueAttribute, value: '' });
+    }
+    onChange(updates);
+  };
+
+  return (
+    <Fieldset
+      label='List of possible Values'
```
weissreto, line 58: Values -> values

packages/new-editor/src/components/editor/variables/detail/Value.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <SimpleSelect
+          value={variable.value}
+          items={filterEnumValuesForSelect()}
+          emptyItem={true}
+          onValueChange={(value: string) => onChange([{ key: treeNodeValueAttribute, value: value }])}
+        />
+      );
+      break;
+    default:
+      input = <Input value={variable.value} onChange={event => onChange([{ key: treeNodeValueAttribute, value: event.target.value }])} />;
```
weissreto, line 46: For passwords display only **** instead of clear text and maybe an eye to display the clear text. Can be done in new PR

weissreto (approved):
Some small comments. Otherwise good

packages/new-editor/src/components/editor/variables/detail/Value.tsx
```diff
@@ -0,0 +1,50 @@
...
+};
+
+export const Value = ({ variable, onChange }: ValueFieldsetProps) => {
+  const filterEnumValuesForSelect = () => {
+    const metadata = asEnumMetadata(variable.metadata);
+    if (!metadata) {
+      return [];
+    }
+    const values = new Set(metadata.values);
+    return Array.from(values)
```
ivy-lli, line 17: ? create a set from an array to create again an array from it?
Do you want to filter duplications here?
If yes, wrong place for this IMO. The table should not allow to create multiple times the same value.

packages/new-editor/src/components/editor/variables/detail/Value.tsx
```diff
@@ -0,0 +1,50 @@
...
+      .filter(value => value !== '')
+      .map(value => {
+        return { label: value, value: value };
+      });
```
ivy-lli, line 21:
```ts
.map(value => ({label: value, value}))
```

packages/new-editor/src/components/editor/variables/detail/Value.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { treeNodeValueAttribute } from '../../../../types/config';
+import { asEnumMetadata, metadataTypeDaytime, metadataTypeEnum, type Variable, type VariableUpdates } from '../../data/Variable';
+
+type ValueFieldsetProps = {
+  variable: Variable;
+  onChange: (updates: VariableUpdates) => void;
+};
+
+export const Value = ({ variable, onChange }: ValueFieldsetProps) => {
+  const filterEnumValuesForSelect = () => {
```
ivy-lli, line 11: why this name? IMO there is no filter functionality here.
This function creates the select items.

In this case I would use a `useMemo` hook. to make sure that this happens only once.

```ts 
const enumValues = useMemo(() => {
  const metadata = asEnumMetadata(variable.metadata);
  if (!metadata) {
    return [];
  }
  return Array.from(values)
    .filter(value => value !== '')
    .map(value => ({label: value, value}));
}, [variable])
```

packages/new-editor/src/components/editor/variables/detail/Value.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const values = new Set(metadata.values);
+    return Array.from(values)
+      .filter(value => value !== '')
+      .map(value => {
+        return { label: value, value: value };
+      });
+  };
+
+  let input;
+  switch (variable.metadata.type) {
```
ivy-lli, line 25: and for this I would create a method:

```ts
const input = (variable, onChange) => {
  switch (variable.metadata.type) {
    case 'daytime': 
      return <Input .../>
  ....
}
```

ivy-lli (changes requested):

packages/new-editor/src/components/editor/variables/detail/Metadata.tsx
ivy-lgi: This makes sense. I guess the only other reason that comes to mind is that this is less error prone regarding typos. I could write `case 'daytine` and not realize it. But this is circumvented as well by writing good tests. I'm gonna change it as you suggested.

ivy-lgi: Actually, I cannot make a typo because TypeScript will catch it.

ivy-lli: That's one big benefit from the typescript system. If you define a Union then typescript will check it for you :)

packages/new-editor/src/components/editor/variables/master/Variables.tsx
ivy-lgi: I moved the structured cloning inside of the methods that actually perform changes to the state. This way, I don't need to pass a clone to these methods every time and I cannot forget to do it in the first place. I also test that these methods do not perform any changes to the data passed to them. See:

https://github.com/axonivy/config-editor-client/blob/c6e8b8fb0a86ab6ae57e44f8c142d2b5b7159b04/packages/new-editor/src/data/data.ts#L28-L36

and

https://github.com/axonivy/config-editor-client/blob/c6e8b8fb0a86ab6ae57e44f8c142d2b5b7159b04/packages/new-editor/src/data/data.test.ts#L75-L92

In this case, `deleteFirstSelectedRow` even returns an object with additional information. So, a switch to the lambda method from `setVariables` wouldn't be that straight forward.

ivy-lli: Ah okey 👍 
Ok for me :)

packages/new-editor/src/components/editor/variables/detail/Variable.tsx
ivy-lgi: Already done in the next PR https://github.com/axonivy/config-editor-client/pull/54

packages/new-editor/src/components/editor/variables/detail/EnumValues.tsx
```diff
@@ -24,7 +24,11 @@ export const PossibleEnumValues = ({ selectedValue: value, values, onChange }: P
   ];
   const meta = {
     updateData: (rowId: string, _columnId: string, value: unknown) => {
```
ivy-lli, line 26: I've changed the updateData interface in on of the latest ui-components updates. 
So the value should now be typed as string. Then you don't have to cast it manually.

ivy-lli (approved):
Thanks for the improvements!

PR #56, opened 2024-04-15
XIVY-11980 implement password input field

ivy-lgi:
I guess the `PasswordInput` component should go live in ui-components?
Also, there's no eye icon in `IvyIcons` as of yet.

password-input


packages/new-editor/src/components/variables/detail/PasswordInput.tsx
```diff
@@ -0,0 +1,20 @@
...
+type PasswordInputProps = {
+  value: string;
+  onChange: (newValue: string) => void;
+};
+
+export const PasswordInput = ({ value, onChange }: PasswordInputProps) => {
+  const [showPassword, setShowPassword] = useState(false);
+
+  return (
+    <div className='password-container'>
```
ivy-lli, line 15: Use the InputGroup component for this :)

ivy-lli (approved):
Yes maybe this should be part of the ui-components. But as long as there is only project which uses this, it is totally fine to let it here until we need it somewhere else.

ivy-lli (approved):

weissreto (approved):
Nice. We will need this component in other config editors too (database, rest, was, etc). But ok for now to include it here.

PR #57, opened 2024-04-15
XIVY-11980 implement variable search

ivy-lgi:
search

### DIFF
--- a/packages/new-editor/src/utils/tree/tree.ts
+++ b/packages/new-editor/src/utils/tree/tree.ts
@@ -96,7 +96,7 @@ export const treeGlobalFilter = <TNode extends TreeNode<TNode>>(data: Array<TNod
       return true;
     }
   }
-  const node = nodesOnPath[nodesOnPath.length - 1];
+  const node = nodesOnPath.at(-1);
   if (node) {
     return node.value.toLowerCase().includes(filterValue);
   }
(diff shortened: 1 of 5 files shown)


weissreto (approved):
👍

packages/new-editor/src/utils/tree/tree.ts
```diff
@@ -79,3 +80,25 @@ const toTreePath = (rowId: string) => {
...
+
+export const treeGlobalFilter = <TNode extends TreeNode<TNode>>(data: Array<TNode>, path: TreePath, filterValue: string) => {
+  filterValue = filterValue.toLowerCase();
+  const nodesOnPath = getNodesOnPath(data, path);
+  for (const node of nodesOnPath) {
+    if (node?.name.toLowerCase().includes(filterValue)) {
+      return true;
+    }
+  }
+  const node = nodesOnPath[nodesOnPath.length - 1];
```
ivy-lli, line 99: if you want the last element of an array you can also do nodesOnPath.at(-1)

ivy-lli (commented):
Why did you need to implement your own filter function?
Not that it is wrong, but as far as I tested it, the default filter function is looking for all columns and is case insensitive.
Or am I wrong?

ivy-lli:
Ah, hmm...
Yes, interesting thought. Not 100% sure if this is for everyone useful, but can also not pretend otherwise.
I think we need to test this first with maybe a very large variable.yaml :)
Maybe we could later also add a setting to change this filter behavior.

Let it like that, then we can get user feedback 👍

ivy-lli (approved):

PR #58, opened 2024-04-17
XIVY-11980 implement fileMetadata

ivy-lgi:
I think this is fine for now but would it make sense to change the value input to a file picker in the future?
I guess this is not so straight-forward to implement.

file-metadata

### DIFF
--- a/packages/new-editor/src/components/variables/data/variable.test.ts
+++ b/packages/new-editor/src/components/variables/data/variable.test.ts
@@ -1,4 +1,12 @@
-import { isEnumMetadata, metadataOptions, toEnumMetadataUpdate, variableMetadataAttribute } from './variable';
+import {
+  fileMetadataFilenameExtensionOptions,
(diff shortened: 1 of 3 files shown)


ivy-lli (approved):
Looks good!
I don't understand why we should need a file picker here. The Variable name is equal to the file path and name.

ivy-lgi:
Maybe it would be easier for customers to choose a file using a file picker and then we convert it to the correct path and name on our side. I think having both options at the same time would be nice but maybe not worth the effort.

ivy-lli:
But only that a file variable exists, doesn't mean that the file also exists. It can also be used to tell the engine where it should write the value for this variable.

weissreto:
I would display the content of the file in the value text field. The text field could then have an icon. If you press on the icon we show a full monaco editor with the content of the file. Very similar to code fields in inscription masks

weissreto (approved):
Looks good

PR #59, opened 2024-04-18
XIVY-11980 accept content as yaml string

ivy-lgi:


weissreto (approved):
Some minor comments and improvements. Overall good.

packages/variable-editor/src/utils/string/string.ts
```diff
@@ -0,0 +1,7 @@
+export const removeLeadingWhitespacesFromEachLine = (string: string) => {
+  return string.replace(/^[^\S\n]+/gm, '');
+};
+
+export const getLastLine = (string: string) => {
+  return string.split('\n').at(-1);
```
ivy-lli, line 6: I think there is a util function in the ui-components splitNewLine or something like that which also consider \r.

PR #60, opened 2024-04-18
Xivy 11980/minor fixes

ivy-fhe:
- Fixed master-variable view to enable scrolling on smaller displays
- Fixed delete-button for last row in master-view

### DIFF
--- a/packages/variable-editor/src/components/variables/master/VariablesMaster.tsx
+++ b/packages/variable-editor/src/components/variables/master/VariablesMaster.tsx
@@ -88,6 +88,9 @@ export const VariablesMaster = ({ variables, setVariables, setSelectedVariablePa
     setSelectedVariablePath([]);
   };
 
+  /* workaround for "table.getIsSomeRowsSelected" as it returns false if only last remaining row is selected */
+  const isRowSelected = () => document.querySelectorAll('[data-state="selected"]').length > 0;
+
   return (
     <Fieldset
       className='variable-wrapper'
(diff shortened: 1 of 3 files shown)


packages/variable-editor/src/components/variables/master/VariablesMaster.tsx
```diff
@@ -88,14 +88,23 @@ export const VariablesMaster = ({ variables, setVariables, setSelectedVariablePa
     setSelectedVariablePath([]);
   };
 
+  /* workaround for "table.getIsSomeRowsSelected" as it returns false if only last remaining row is selected */
+  const isRowSelected = () => document.querySelectorAll('[data-state="selected"]').length > 0;
```
ivy-lgi, line 92: I think you can do it easier with something like `Object.keys(table.getState().rowSelection).length > 0`

ivy-lgi (approved):
Check if my suggestion suits you, otherwise LGTM :+1:

ivy-lgi:
@ivy-lli Is there no scrolling functionality in ui-components?

packages/variable-editor/src/components/variables/master/VariablesMaster.tsx
ivy-lgi: I also think you should move it to `utils/table/table.ts` as it is a general table functionality and can potentially be reused. Then, you can also test it :wink:

ivy-fhe: :+1: was not aware this function exists

ivy-lli:
I think this is something each adapter has to solve for its own... Not every editor has the same requirement for scrolling, sometimes the table needs to scroll sometimes the whole master part.

If you mean something like a scrollable body, where the table header is sticky then the answer is no, but this I can imagine that it would be cool to have in the ui-components.

PR #61, opened 2024-04-19
XIVY-11980 parse variables to yaml

ivy-lgi:
Figured it out :smile:


packages/variable-editor/src/components/variables/data/variable-utils.ts
```diff
@@ -125,3 +133,74 @@ const enrichVariableWithFileMetadata = (variable: Variable, metadata: string) =>
...
+    variableValue = parseVariables(variable.children);
+  }
+  return new Pair(variableKey, variableValue);
+};
+
+const parseKeyDescription = (variable: Variable) => {
+  const metadata = variable.metadata;
+  const description = variable.description;
+  if (metadata.type === '') {
+    return description && addSingleLeadingWhitespaceToEachLine(description);
```
weissreto, line 163: This seems to be wrong?

packages/variable-editor/src/components/variables/data/variable-utils.ts
```diff
@@ -125,3 +133,74 @@ const enrichVariableWithFileMetadata = (variable: Variable, metadata: string) =>
...
+const parseMetadataComment = (metadata: Metadata) => {
+  let metadataComment;
+  switch (metadata.type) {
+    case 'enum':
+      if (isEnumMetadata(metadata)) {
+        metadataComment = 'enum: ' + metadata.values.join(', ');
+      }
+      break;
+    case 'file':
+      if (isFileMetadata(metadata)) {
```
weissreto, line 180: Same here

packages/variable-editor/src/components/variables/data/variable-utils.test.ts
```diff
@@ -14,84 +27,82 @@ describe('toVariables', () => {
...
-                metadata: { type: '' },
-                children: []
-              }
-            ]
-          }
-        ]
-      }
-    ]);
-  });
+  variables = [
```
weissreto, line 30: Hard to read this tests with all these test data in it. Can we move them to individual files and import them?

weissreto (approved):
😎🙏. Some minor comments

packages/variable-editor/src/components/variables/data/variable-utils.ts
ivy-lgi: This is a shortcut for:
```
    if (description) {
      return addSingleLeadingWhitespaceToEachLine(description);
    }
    return;  
```
or even more verbose:
```
    if (description !== '') {
      return addSingleLeadingWhitespaceToEachLine(description);
    }
    return undefined;
```

Gonna change it to make it more readable.

weissreto: Ah I though it is a string concatenation. Sorry my fault that I don't know type script. All good.

packages/variable-editor/src/components/variables/data/variable-utils.ts
weissreto: Ahh, sorry I had a comment on line 178 which was swallowed. Why are you double checking if the type is file or enum? Isn't it clear with the switch already?

weissreto (approved):

packages/variable-editor/src/components/variables/data/variable-utils.ts
ivy-lgi: This is a type guard function:
```
isFileMetadata = (metadata?: Metadata): metadata is FileMetadata => {...}
```
You can see that it has `metadata is FileMetadata` in its signature. If this function returns true, TypeScript accepts that the argument is of type `FileMetadata` within the scope of the if block. Otherwise, TypeScript would show an error on the next line when trying to access the attribute `values`, since it does not exist on the type `Metadata` but only on its sub-type `FileMetadata`. Essentially, its a save way to cast an object of type `Metadata` to its sub-type `FileMetadata`.

weissreto: Understand. good

ivy-lli (commented):
Quick scan. looks good 👍

PR #71, opened 2024-04-25
XIVY-13805 communication for old designer

ivy-lgi:


ivy-lgi:
https://github.com/axonivy/core/pull/5646

ivy-lgi:
- Added communication for old designer in [/packages/variable-editor/src/protocol](https://github.com/axonivy/config-editor-client/pull/71/files#diff-4cd30040dafa5e079cef7b3f434a3d332d46d69b52c340b26b7db32c81cf9791)
- Adjusted [VariableEditor](https://github.com/axonivy/config-editor-client/pull/71/files#diff-241ecee69a3d39a2c78826b09a42129d6fee38b740ae06d3e311fce7bebb22a1) to use communication
- Adjusted [launch config](https://github.com/axonivy/config-editor-client/pull/71/files#diff-bd5430ee7c51dc892a67b3f2829d1f5b6d223f0fd48b82322cfd45baf9f5e945) to work with a running designer and added launch config with [mock client](https://github.com/axonivy/config-editor-client/pull/71/files#diff-e8a692fa442aa93512f7db64228a4ac76258c11a2a78c9a7a5fa1dca51199d37) for easier local development

integrations/standalone/vite.config.ts
```diff
@@ -9,7 +9,12 @@ export default defineConfig({
     chunkSizeWarningLimit: 5000,
     rollupOptions: { input: { index: './index.html' } }
   },
-  server: { port: 3000 },
+  esbuild: {
+    supported: {
+      'top-level-await': true
+    }
```
ivy-lli, line 15: what is this?

packages/variable-editor/src/protocol/ClientContextProvider.tsx
```diff
@@ -0,0 +1,15 @@
+import type { ReactNode } from 'react';
+import { createContext, useContext } from 'react';
+import type { Client, ClientContext } from './types';
+
+const defaultClientContext: any = undefined;
+const ClientContextInstance = createContext<ClientContext<any>>(defaultClientContext);
+
+export const useClient = (): Client<any> => {
```
ivy-lli, line 8: why do you need to use any here?

.vscode/launch.json
```diff
@@ -5,10 +5,18 @@
   "version": "0.2.0",
   "configurations": [
     {
-      "name": "Run Standalone (Chrome)",
+      "name": "Launch Standalone (Chrome)",
       "type": "chrome",
       "request": "launch",
-      "url": "http://localhost:3000",
+      "url": "http://localhost:3001/?server=localhost:8081&app=designer&pmv=variables&file=/variables.yaml",
```
ivy-lli, line 11: I think the file is useless as a parameter. This editor always works with the variables.yaml file. And there exists only one variables.yaml file per pmv...

I think it is ok to have this in the lsp protocol (as we maybe can share it for all config editors), but I would hard code it for this editor.

Or did you have a special thing in thought that you added it here?

integrations/standalone/vite.config.ts
ivy-lgi: Without this, packaging will throw the following error in `config-editor-client/integrations/standalone/src/index.tsx`:
```
@axonivy/config-editor-standalone: Error: [vite:esbuild-transpile] Transform failed with 1 error:
@axonivy/config-editor-standalone: assets/index-!~{001}~.js:25422:15: ERROR: Top-level await is not available in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14" + 2 overrides)
@axonivy/config-editor-standalone: 
@axonivy/config-editor-standalone: Top-level await is not available in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14" + 2 overrides)
@axonivy/config-editor-standalone: 25420|  const file = URLParams.file();
@axonivy/config-editor-standalone: 25421|  const theme = URLParams.theme();
@axonivy/config-editor-standalone: 25422|  const client = await ClientJsonRpc.startWebSocketClient(server);
@axonivy/config-editor-standalone:    |                 ^
```
Is there a better way to handle this?

packages/variable-editor/src/protocol/ClientContextProvider.tsx
ivy-lgi: Because this is a `.tsx` file, it seems to misinterpret type arguments:
image
I could move this function to a different file though if you think that's better.

.vscode/launch.json
ivy-lgi: You are right, it is not necessary right now. I'm not sure how we are going to handle multiple config editors living here yet but what I was thinking is that once we have different editors we can identify the editor to use with this argument.
```
root.render(
  <React.StrictMode>
    <ThemeProvider defaultTheme={theme}>
      <ClientContextProvider client={client}>
        <QueryProvider client={queryClient}>
          {file === 'variables.yaml' && <VariableEditor app={app} pmv={pmv} file='variables.yaml' />}
          {file === 'roles.xml' && <RolesEditor app={app} pmv={pmv} file='roles.xml' />}
        </QueryProvider>
      </ClientContextProvider>
    </ThemeProvider>
  </React.StrictMode>
);
```
Then, we probably will have multiple launch configs with the different files in the url here. But I can remove it for now if you want me to.

ivy-lli: I would remove it for now, we can add it later again if we want to

packages/variable-editor/src/protocol/ClientContextProvider.tsx
ivy-lli: add a comma: export const useClient = <T,>():

integrations/standalone/vite.config.ts
ivy-lli: As the other editors do it (where I assume you copied it ;)):
Put it into an async function which you then simply call.

```tsx
export async function start(): Promise<void> {
  const server = URLParams.webSocketBase();
  const app = URLParams.app();
  const pmv = URLParams.pmv();
  const file = URLParams.file();
  const theme = URLParams.theme();

  const client = await FormClientJsonRpc.startWebSocketClient(server);
  const queryClient = initQueryClient();

  createRoot(document.getElementById('root')!).render(
    <React.StrictMode>
      <ThemeProvider defaultTheme={theme}>
        <ClientContextProvider client={client}>
          <QueryProvider client={queryClient}>
            <App app={app} pmv={pmv} file={file} />
          </QueryProvider>
        </ClientContextProvider>
      </ThemeProvider>
    </React.StrictMode>
  );
}

start();
```

packages/variable-editor/src/protocol/ClientContextProvider.tsx
```diff
@@ -5,11 +5,11 @@ import type { Client, ClientContext } from './types';
 const defaultClientContext: any = undefined;
 const ClientContextInstance = createContext<ClientContext<any>>(defaultClientContext);
```
ivy-lli, line 6: still `any` here.
why can't you directly here define that this is a string context? Then the generic stuff wouldn't even be necessary...

ivy-lgi: I defined the data that is being sent/received as a generic, as it will include different kinds of meta data for different kinds of editors and therefore be of a different type.
`export type Data<T> = { context: DataContext; data: T };`

But yes, for now it is a simple string in any case, so I could simplify it for now.

ivy-lli: I'm always a fan of keeping things in mind which may come later, but that doesn't mean that the code needs to show these cases already. 
IMO it is easier to make these changes when (in this case) a second editor will be implemented. Because then we know better what we really need.

Maybe we never share this context overall editors, and then we made this generic stuff here but never used it...

PR #83, opened 2024-05-14
XIVY-14079 focus name input of new variable

ivy-lgi:
Not sure why but this actually does work and really only takes focus after adding a new variable.

focus-name-input


ivy-lli (approved):
Fine solution for me. If this is not enough we can improve it more later :) 👍

PR #86, opened 2024-05-16
improve variables parsing > master

ivy-lgi:
- Comments at the start and end of the `variables.yaml` file will be preserved
- Improved handling of empty and malformed `variables.yaml`

What does not work is adding a comment after the last variable in a mapping:
```
Variables:
  someConnector:
    someKey: someValue
    # comment after last variable in mapping
  # comment before next mapping
  anotherConnector:
    anotherKey: anotherValue
```

These two comments will be combined into one comment of `someConnector` by the library and removed by the variables editor as of now. I don't see a way to differentiate some possible cases. I will add a warning about manual changes to the `variables.yaml` in the documentation.


packages/variable-editor/src/VariableEditor.tsx
```diff
@@ -60,14 +60,16 @@ function VariableEditor(props: DataContext) {
     return <p>{'An error has occurred: ' + error}</p>;
   }
 
-  const variables = toVariables(data.data);
+  const rootVariable = toVariables(data.data);
```
weissreto, line 63: How many times have you renamed this variable from 'variables' to 'rootVariable' and back again? ;-)

weissreto (approved):
Looks good.

packages/variable-editor/src/VariableEditor.tsx
ivy-lgi: I don't think I have, maybe you mean another one? :sweat_smile:

ivy-lli (approved):

PR #92, opened 2024-05-27
run update axonivy next before building & manual package lock fix

ivy-lmu:
... as we do it in other repos


ivy-lmu:
same problem as here https://github.com/axonivy/form-editor-client/pull/60

ivy-cst (approved):
:+1:

PR #94, opened 2024-05-28
XIVY-13802 get validation messages

ivy-lgi:
Simply displaying the messages at the top for now.
image


ivy-lli (approved):
Nice.
I assume the path is currently wrong or you need to map it to the correct row right?

ivy-lgi:
The path is correct, I just need to throw it on the correct variable and think about how to display it nicely.

ivy-lli:
> The path is correct, I just need to throw it on the correct variable and think about how to display it nicely.

If possible I would reuse the message row for this

PR #102, opened 2024-06-10
fix console warning

ivy-lgi:
"Query data cannot be undefined. Please make sure to return a value other than undefined from your query function. Affected query key: ["validate"]"


ivy-lli:
Technically, instead of hold the validations inside it's own state, you could also just use the data from the validation query request and from the save data just update this query data (like the inscription does it https://github.com/axonivy/inscription-client/blob/master/packages/editor/src/App.tsx#L77)

But ok for me

PR #103, opened 2024-06-11
XIVY-14003 beautify empty detail view

ivy-lgi:
Target:
image

Result:
beautiful-detail-view

### DIFF
--- a/packages/variable-editor/src/components/detail/EmptyDetail.css
+++ b/packages/variable-editor/src/components/detail/EmptyDetail.css
@@ -1,7 +1,5 @@
 .empty-detail {
   flex: auto;
-  align-items: center;
-  justify-content: center;
   margin: 0 20%;
   color: var(--N800);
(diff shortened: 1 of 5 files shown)


ivy-lgi:
Correct icon is not available yet, need to ask Anaïs.

packages/variable-editor/src/components/detail/EmptyDetail.css
```diff
@@ -0,0 +1,16 @@
+.empty-detail {
+  flex: auto;
+  align-items: center;
+  justify-content: center;
```
ivy-lli, line 4: these two are also available as properties on the flex component, if you like :)

ivy-lli (approved):
Cool :)

ivy-lli:
Maybe this empty view could also be something for the ui-components, as we may could use it also for the other editors :)

PR #108, opened 2024-06-13
Feature/XIVY-14000 add playwright test for master view

weissreto:
Add tests for add, delete variable, collapse, exand tree, password, theme

Add page objects for VariableEditor, Settings, Theme, Table, Switch, Button, TextArea


ivy-lgi (approved):
Very cool :sunglasses:
Good for now but maybe some tests could be more rigorous? Check my comments and add what you think makes sense.

PR #110, opened 2024-06-14
Feature/XIVY-14000 add playwright test for detail view

weissreto:
Add test for variable details view

### DIFF
--- a/integrations/standalone/tests/mock/variable-editor-detail.spec.ts
+++ b/integrations/standalone/tests/mock/variable-editor-detail.spec.ts
@@ -85,4 +85,51 @@ test.describe('VariableEditor Detail', () => {
     await details.expectValues('myName', 'test.txt', 'This is myName with a value of myValue', 'File');
     await details.fileNameExtension.expectValue('txt');
   });
+
+  test('new enum variable', async () => {
+    await editor.tree.row(0).click();
+    await editor.add.click();
+    await editor.tree.row(11).click();
+    const details = editor.details;
+    await details.expectTitle('Variables Editor - ');
+
+    await details.fill('myName', 'Monday', 'This is myName with a value of Monday', 'Enum');
+
+    await details.expectTitle('Variables Editor - myName');
+    await editor.tree.row(11).expectValues(['myName', 'Monday']);
+    await editor.tree.row(10).click();
+    await details.expectTitle('Variables Editor - connectorProvider');
+    await editor.tree.row(11).click();
+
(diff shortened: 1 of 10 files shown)


integrations/standalone/tests/mock/variable-editor-detail.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+  });
+
+  test('new password variable', async () => {
+    await editor.tree.row(0).click();
+    await editor.add.click();
+    await editor.tree.row(11).click();
+    const details = editor.details;
+    await details.expectTitle('Variables Editor - ');
+
+    await details.fill('myName', 'myValue', 'This is myName with a value of myValue', 'Password');
```
ivy-lgi, line 37: After changing the metadata type to `Password`, maybe we should also check that the value input field now has the type `password`?

integrations/standalone/tests/mock/variable-editor-detail.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+  });
+
+  test('new daytime variable', async () => {
+    await editor.tree.row(0).click();
+    await editor.add.click();
+    await editor.tree.row(11).click();
+    const details = editor.details;
+    await details.expectTitle('Variables Editor - ');
+
+    await details.fill('myName', '12:15', 'This is myName with a value of myValue', 'Daytime');
```
ivy-lgi, line 56: After changing the metadata type to `Daytime`, maybe we should also check that the value input field now has the type `daytime`?

integrations/standalone/tests/mock/variable-editor-detail.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+    await details.enumValue.choose('Wednesday');
+    await details.enumValue.expectValue('Wednesday');
+
+    await details.enumValue.choose('Monday');
+
+    await details.listOfPossibleValues.deleteValue('Tuesday');
+    await details.listOfPossibleValues.deleteValue('Wednesday');
+
+    await details.listOfPossibleValues.expectValues(['Monday']);
+  });
```
ivy-lgi, line 133: There are some edge cases here we had to handle that might want to get tested?
- What If you delete the element currently selected as the value? -> I think the value becomes empty.
- What if you add the same value twice? -> I think it works but still only shows up once in the value selection.

ivy-lgi (approved):
Maybe some things to add, otherwise good :+1:

PR #112, opened 2024-06-17
XIVY-14003 change password icon

ivy-lgi:
change password icon
update internal dependencies

old:
image

new:
image


weissreto (commented):
👍

PR #114, opened 2024-06-17
XIVY-14000 Add engine integration tests

weissreto:


ivy-lgi (approved):
Very cool :sunglasses: 
One question: Do I understand correctly that in the integration tests only the validation is tested? Shouldn't everything that is tested with the mock also be tested in the integration tests?

integrations/standalone/tests/integration/project/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>ch.ivyteam.variable</groupId>
+  <artifactId>integration</artifactId>
+  <version>11.4.0-SNAPSHOT</version>
+  <packaging>iar-integration-test</packaging>
+
+  <properties>
+    <engine.page.url>https://product.ivyteam.io/</engine.page.url>
+    <project.build.plugin.version>11.4.0-SNAPSHOT</project.build.plugin.version>
+    <engine.test.app>inscription-test</engine.test.app>
```
ivy-lli, line 13: it does make no difference, but maybe adjust this name here :)

integrations/standalone/tests/integration/project/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>ch.ivyteam.variable</groupId>
+  <artifactId>integration</artifactId>
+  <version>11.4.0-SNAPSHOT</version>
+  <packaging>iar-integration-test</packaging>
+
+  <properties>
+    <engine.page.url>https://product.ivyteam.io/</engine.page.url>
+    <project.build.plugin.version>11.4.0-SNAPSHOT</project.build.plugin.version>
```
ivy-lli, line 12: as we now have an ivy project in this repo, we should also adjust our raise version scripts:
- add raise-build-plugin-version.sh
- add this repo to the raise project build plugin version job
- adjust the raise-version.sh script to also update this project version

weissreto:
> Very cool 😎 One question: Do I understand correctly that in the integration tests only the validation is tested? Shouldn't everything that is tested with the mock also be tested in the integration tests?

Is there a chance that a feature won't run if we integrate with an engine rather than the mock?

weissreto:
I think it is good enough if we test certain things with mock only and other things in the integration. The target is not to have a 100% code coverage but to ensure that nothing important is broken. And I think with the unit, mock and integration test we have a good test coverage now that we find out pretty fast if a feature is broken.

ivy-lli:
> Very cool 😎 One question: Do I understand correctly that in the integration tests only the validation is tested? Shouldn't everything that is tested with the mock also be tested in the integration tests?

Definilty not! otherwise we can delete the mock test...
The integration tests should only test the editor can connect to our backend and this communication work and the server delivers the info which the editor expect.
Everything else can we also check without a backend.
Remember that the mock test are not the fastest but still much faster than integration tests and much stabler, as they don't depend on an other source!

ivy-lgi:
> > Very cool 😎 One question: Do I understand correctly that in the integration tests only the validation is tested? Shouldn't everything that is tested with the mock also be tested in the integration tests?
> 
> Is there a chance that a feature won't run if we integrate with an engine rather than the mock?

The only thing I think is the communication itself.
The validation communication is now being tested with this new integration test.

The acquiring of data and the saving of data is not explicitly tested.
Maybe we should add simple tests to check that the Variables Editor shows the data initially present in the `variables.yaml` and saves the changes done with it to the `variables.yaml`.

Then, all the lsp endpoints are tested.

weissreto:
> > > Very cool 😎 One question: Do I understand correctly that in the integration tests only the validation is tested? Shouldn't everything that is tested with the mock also be tested in the integration tests?
> > 
> > 
> > Is there a chance that a feature won't run if we integrate with an engine rather than the mock?
> 
> The only thing I think is the communication itself. The validation communication is now being tested with this new integration test.
> 
> The acquiring of data and the saving of data is not explicitly tested. Maybe we should add simple tests to check that the Variables Editor shows the data initially present in the `variables.yaml` and saves the changes done with it to the `variables.yaml`.
> 
> Then, all the lsp endpoints are tested.

Ok, I will add that. But in a different PR.

ivy-lli:
> > > Very cool 😎 One question: Do I understand correctly that in the integration tests only the validation is tested? Shouldn't everything that is tested with the mock also be tested in the integration tests?
> > 
> > 
> > Is there a chance that a feature won't run if we integrate with an engine rather than the mock?
> 
> The only thing I think is the communication itself. The validation communication is now being tested with this new integration test.
> 
> The acquiring of data and the saving of data is not explicitly tested. Maybe we should add simple tests to check that the Variables Editor shows the data initially present in the `variables.yaml` and saves the changes done with it to the `variables.yaml`.
> 
> Then, all the lsp endpoints are tested.

Yes I think it is ok to add some tests for this. Keep in mind that:
1. the lsp itself is also tested in the core (and IMO it is much easier to test it there). But some basic test are definilty good
2. it is not so easy to directly read the written content of the variables.yaml, maybe it is just much easier to reload the page after the save to check that the new data is delivered back :)

ivy-lli (approved):
thx 👍

PR #122, opened 2024-06-24
XIVY-14264 use key to match validation messages

ivy-lgi:
Core change: https://github.com/axonivy/core/pull/6217

### DIFF
--- a/packages/variable-editor/src/components/variables/data/test-utils/types.ts
+++ b/packages/variable-editor/src/components/variables/data/test-utils/types.ts
@@ -1,0 +1,60 @@
+import { type Cell, type FilterMeta, type Row, type RowPinningPosition } from '@tanstack/react-table';
+
+export class MockRow<TData> implements Row<TData> {
+  original;
+  parentRows;
+
+  constructor(original: TData, parentRows: Array<MockRow<TData>>) {
+    this.original = original;
+    this.parentRows = parentRows;
+  }
+
+  getParentRows = () => {
+    return this.parentRows;
+  };
+
+  _getAllCellsByColumnId: () => Record<string, Cell<TData, unknown>>;
+  _uniqueValuesCache: Record<string, unknown>;
+  _valuesCache: Record<string, unknown>;
+  depth: number;
+  getAllCells: () => Cell<TData, unknown>[];
+  getLeafRows: () => Row<TData>[];
+  getParentRow: () => Row<TData> | undefined;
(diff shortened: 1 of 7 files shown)


packages/variable-editor/src/components/variables/data/test-utils/types.ts
```diff
@@ -0,0 +1,60 @@
+import { type Cell, type FilterMeta, type Row, type RowPinningPosition } from '@tanstack/react-table';
+
+export class MockRow<TData> implements Row<TData> {
```
ivy-rew, line 3: Maybe someone with more typescript knowledge than I have is able to simplify this mock definition, so that we only need to mock out the parts we're actually programming against?

ivy-rew (approved):

packages/variable-editor/src/components/variables/data/validation-utils.test.ts
```diff
@@ -1,49 +1,55 @@
 import type { ValidationMessages } from '../../../protocol/types';
+import { MockRow } from './test-utils/types';
 import { containsError, containsWarning, toValidationMessageVariant, validationMessagesOfRow } from './validation-utils';
+import type { Variable } from './variable';
 
 let validationMessages: ValidationMessages;
+let rowWithMessages: MockRow<Variable>;
+let rowWithoutMessages: MockRow<Variable>;
```
ivy-lli, line 8: why create this variables here, but add the values below?

packages/variable-editor/src/components/variables/data/test-utils/types.ts
ivy-lli: exactly! not ideal, otherwise if we update tanstack table and a new property gets added this will fail.

Also most of the times classes should not be used in js just use types and objects.
if you need a creator then do a creator function which returns a row object.
e.g. like here: https://github.com/axonivy/inscription-client/blob/master/packages/editor/src/components/browser/function/parent-name.test.ts

ivy-lli (changes requested):

packages/variable-editor/src/components/variables/data/validation-utils.test.ts
ivy-lgi: I saw this being done this way in other tests. I assumed it may have something to do with how the tests work or to guarantee that this data can never change between tests.

Can I instantiate these right here?

ivy-lli: I would say so, expect they change...

PR #128, opened 2024-07-02
XIVY-14357 add options of all known namespaces

ivy-lgi:
any known namespace can now be selected in the combobox
a new namespace can be input as well

added unittests for gathering all known namespaces
improved playwright test to check that known namespaces are present in the selection of the combobox

### DIFF
--- a/packages/variable-editor/src/components/variables/master/AddVariableDialog.css
+++ b/packages/variable-editor/src/components/variables/master/AddVariableDialog.css
@@ -1,3 +1,11 @@
 .add-variable-dialog-trigger-button {
   margin-left: auto;
 }
+
+.ui-combobox-menu {
+  z-index: 50;
+}
+
+.ui-combobox-menu:empty {
+  display: none;
+}
--- a/packages/variable-editor/src/components/variables/master/AddVariableDialog.css
+++ b/packages/variable-editor/src/components/variables/master/AddVariableDialog.css
@@ -2,10 +2,6 @@
   margin-left: auto;
 }
 
-.ui-combobox-menu {
-  z-index: 50;
-}
-
 .ui-combobox-menu:empty {
   display: none;
 }
--- a/packages/variable-editor/src/components/variables/master/AddVariableDialog.tsx
+++ b/packages/variable-editor/src/components/variables/master/AddVariableDialog.tsx
@@ -15,7 +15,7 @@ import {
 import { IvyIcons } from '@axonivy/ui-icons';
 import { type Table } from '@tanstack/react-table';
 import { useState } from 'react';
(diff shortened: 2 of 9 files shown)


ivy-lgi:
namespace-options

packages/variable-editor/src/components/variables/master/AddVariableDialog.css
```diff
@@ -1,3 +1,11 @@
 .add-variable-dialog-trigger-button {
   margin-left: auto;
 }
+
+.ui-combobox-menu {
+  z-index: 50;
+}
```
ivy-lli, line 7: I think this should not be needed any longer. 
Try to update to the latest ui-components.

packages/variable-editor/src/components/variables/master/AddVariableDialog.css
```diff
@@ -1,3 +1,11 @@
...
+
+.ui-combobox-menu:empty {
+  display: none;
+}
```
ivy-lli, line 11: needed?

packages/variable-editor/src/components/variables/master/AddVariableDialog.tsx
```diff
@@ -73,7 +79,7 @@ export const AddVariableDialog = ({ table, variables, setVariables, setSelectedV
             <Input />
```
ivy-lli, line 79: Maybe not part of this pr, but a very good gif. We should not allow an empty name, but show a message if someone tries to add a variable with no name

packages/variable-editor/src/components/variables/master/AddVariableDialog.css
ivy-lgi: You are right, but I notice that after updating the toolbar also stays while the dialog is open :sweat_smile: 
image

packages/variable-editor/src/components/variables/master/AddVariableDialog.css
ivy-lgi: Not necessarily, but I think it works better. If I remove it, things like this seem cumbersome:
hide-empty-menu

packages/variable-editor/src/components/variables/master/AddVariableDialog.tsx
ivy-lgi: Going to add validation with another PR :+1:

packages/variable-editor/src/components/variables/master/AddVariableDialog.css
ivy-lli: :(
sadly this has also a zindex... Need to remove this. Hopefully which no side effect :D

ivy-lli (approved):

PR #132, opened 2024-07-04
XIVY-14357 validate name in add variable dialog

ivy-lgi:
name cannot be empty or already present in the provided namespace

validation updates on changes made to the name input field or the namespace combobox

create button is disabled if any validation messages are present

added unittests and playwright tests including new pageobjects

name-validation

### DIFF
--- a/packages/variable-editor/src/components/variables/master/AddVariableDialog.tsx
+++ b/packages/variable-editor/src/components/variables/master/AddVariableDialog.tsx
@@ -15,7 +15,7 @@ import {
 } from '@axonivy/ui-components';
 import { IvyIcons } from '@axonivy/ui-icons';
 import { type Table } from '@tanstack/react-table';
-import { useMemo, useState, type ChangeEvent, type FormEvent } from 'react';
+import { useMemo, useState } from 'react';
 import {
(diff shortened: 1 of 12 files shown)


packages/variable-editor/src/components/variables/master/AddVariableDialog.tsx
```diff
@@ -28,14 +36,32 @@ type AddVariableDialogProps = {
...
 
-  const newVariableName = () => {
-    return newNodeName(table, 'NewVariable');
+  const handleNamespaceComboboxChange = (value: string) => {
+    setNamespace(value);
+  };
+
+  const handleNamespaceComboboxInput = (event: FormEvent<HTMLInputElement>) => {
+    setNamespace(event.currentTarget.value);
   };
```
ivy-lli, line 65: why should this function need to know that it will be used for a combobox? it is here to change the namespace...
IMO the first function is anyways completely senseless just call setNamespace on the onChange of the combobox. Not every thing needs to be abstracted into an own function.

ivy-lli (changes requested):

ivy-lli (approved):
thx for the improvements 👍 
Nice feature :)

PR #135, opened 2024-07-04
XIVY-14357 implement variable creation

ivy-lgi:
variable will now be created where the data of the add dialog suggests

- if the namespace is empty, it will be created on the topmost level
- if the namespace exists already, it will be added to that namespace
- if any or all parts of the namespace do not exist already, they will be created as well

with this, the story is finished

create-variable

### DIFF
--- a/integrations/standalone/tests/pageobjects/Combobox.ts
+++ b/integrations/standalone/tests/pageobjects/Combobox.ts
@@ -19,15 +19,12 @@ export class Combobox {
 
   async fill(value: string) {
     await this.locator.fill(value);
+    await this.locator.blur();
   }
 
   async choose(value: string) {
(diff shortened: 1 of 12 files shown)


integrations/standalone/tests/pageobjects/Combobox.ts
```diff
@@ -26,6 +26,10 @@ export class Combobox {
     await this.page.getByRole('option', { name: value }).first().click();
   }
 
+  async blur() {
+    this.locator.blur();
```
ivy-lli, line 30: maybe this should be done after fill always?

ivy-lli (approved):

ivy-lli (approved):

weissreto (commented):
👍

PR #142, opened 2024-07-17
Feature/XIVY-14263 Import and overwrite known variables from required projects

weissreto:
grafik


packages/variable-editor/src/components/variables/master/OverwriteDialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+};
+
+export const OverwriteDialog = ({ knownVariables, table, variables, setVariables, setSelectedVariablePath }: OverwriteProps) => {
+  const toNode = (node: ProjectVarNode): BrowserNode => {
+    const c = node.children.map(child => toNode(child));
+    const icon = nodeIcon(node);
+    const info = node.description;
+    return {
+      value: node.name,
+      info: info,
```
ivy-lmu, line 39: detail: no need to do the mapping if constant name is the same as the field name. e.g. `info,` instead of ?`info: info, `, would be enough here

packages/variable-editor/src/components/variables/master/OverwriteDialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+    }
+  };
+
+  const VariableBrowser = ({ applyFn }: { applyFn: (node?: ProjectVarNode) => void }) => {
+    const variableBrowser = useBrowser(nodes);
+    return (
+      <BrowsersView
+        browsers={[
+          {
+            name: 'Variables',
```
ivy-lmu, line 87: is it somehow possible to hide the `Variables` tab on top of the search? IMO it gives no more info.

ivy-lmu (approved):
nice one 👍

packages/variable-editor/src/components/variables/master/OverwriteDialog.tsx
weissreto: I have added a note with a feature request here https://1ivy.atlassian.net/wiki/spaces/XIVY/pages/245137413/UI+Improvements to support to hide the browser tabs.

packages/variable-editor/src/components/variables/master/OverwriteDialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const nodes = toNodes();
+
+  const insertVariable = (node?: ProjectVarNode): void => {
+    if (node) {
+      const lastDot = node?.key.lastIndexOf('.');
+      const namespace = node?.key.substring(0, lastDot);
+      let metadataType: MetadataType = '';
+      if (isMetadataType(node?.type)) {
+        metadataType = node?.type;
+      }
```
ivy-lgi, line 62: For me, in such a case a ternary operator is more readable:
```
const metadataType = isMetadataType(node.type) ? node.type : '';
```

packages/variable-editor/src/VariableEditor.tsx
```diff
@@ -30,7 +30,8 @@ function VariableEditor(props: EditorProps) {
     return {
       data: () => ['variable-editor', 'data', context],
       saveData: () => ['saveData'],
-      validate: () => ['validate']
+      validate: () => ['validate'],
+      knownVariables: () => ['meta/knownVariables']
```
ivy-lli, line 34: we should add here more info the query key, otherwise, this will maybe clash in the global neo react query context. E.g. like here https://github.com/axonivy/form-editor-client/blob/master/packages/editor/src/components/editor/Editor.tsx#L34-L35

packages/variable-editor/src/protocol/types.ts
```diff
@@ -1,15 +1,28 @@
...
-export interface RequestTypes {
+export interface MetaRequestTypes {
+  'meta/knownVariables': [DataContext, ProjectVarNode];
+}
+
+export interface RequestTypes extends MetaRequestTypes {
   data: [any, any];
   saveData: [any, any];
   validate: [any, any];
+  overwritables: [any, any];
```
ivy-lli, line 25: still needed?

packages/variable-editor/src/protocol/types.ts
```diff
@@ -28,6 +41,7 @@ export interface Client {
   data(context: DataContext): Promise<Data>;
   saveData(saveArgs: SaveArgs): Promise<ValidationMessages>;
   validate(validate: DataContext): Promise<ValidationMessages>;
+  meta<TMeta extends keyof MetaRequestTypes>(path: TMeta, args: MetaRequestTypes[TMeta][0]): Promise<MetaRequestTypes[TMeta][1]>;
```
ivy-lli, line 44: I also think that this one here is not needed.

packages/variable-editor/src/VariableEditor.tsx
```diff
@@ -97,6 +103,7 @@ function VariableEditor(props: EditorProps) {
           setVariables={setVariables}
           setSelectedVariablePath={setSelectedVariablePath}
           validationMessages={validationMessages}
+          knownVariables={knownVariables}
```
ivy-lli, line 106: why do you request fetch this data here and not where it is needed?
This approach has 2 disadvantaches:
- The data is initially loaded for everyone which opens this editor even if he doesn't open this browser at all.
- You need to add properties to the VariablesMaster component, which blows up the interface itself

ivy-lgi (commented):
Some notes from my side:
- The available variables are not updated if the variables editor is already open and a new project is added as a dependency. The editor needs to be closed and reopened. Can we fetch the variables when the dialog opens?
new-dependency

- It is not possible to choose multiple variables at once. I think at least when selecting a folder all subvariables and subfolders should be added as well instead of just the folder. At best, one could select any collection of variables using something like CTRL+Click.
add-multiple

- The dialog box seems kind of small to present all this information. Can we make it bigger?
- The name and description of a variable is shown in the tree itself and in the info box. I think it should only be shown once. This would also increase the available space in the dialog box.
image

- The search is not working as I expected. I can only search for toplevel names. Can we reuse the existing search function of the tree somehow?
search

It works fine and this dialog won't be used often so you can implement/change what you deem feasible. :+1: Maybe add the rest to the [UI Improvements](https://1ivy.atlassian.net/wiki/spaces/XIVY/pages/245137413/UI+Improvements).

ivy-lli:
> Some notes from my side:
> 
>     * The available variables are not updated if the variables editor is already open and a new project is added as a dependency. The editor needs to be closed and reopened. Can we fetch the variables when the dialog opens?

Should be fixed if the data fetch happens inside the browser component and not in the root editor -> see my comment above.

>     * It is not possible to choose multiple variables at once. I think at least when selecting a folder all subvariables and subfolders should be added as well instead of just the folder. At best, one could select any collection of variables using something like CTRL+Click.

We never did a multi select before, but it should be possible. But I'm not sure if it is worth the effort, but I think too select the parent would be a nice approvement.

>     * The dialog box seems kind of small to present all this information. Can we make it bigger?

I think this depends on the height of the editor...

>     * The name and description of a variable is shown in the tree itself and in the info box. I think it should only be shown once. This would also increase the available space in the dialog box.

What if the description is 10 lines? Then this will not show the full description in the row.
If you don't like the info box, just close it...
 
>     * The search is not working as I expected. I can only search for toplevel names. Can we reuse the existing search function of the tree somehow?

Yes this is a problem of the current browser, and yes maybe the ui-component can here provide some filter functions, e.g. the one you have for the variables.

packages/variable-editor/src/protocol/types.ts
weissreto: No

packages/variable-editor/src/protocol/types.ts
weissreto: Do not understand why it is not needed. Can you explain?

ivy-lli: ah sorry, I think I didn't read this correctly :)

PR #147, opened 2024-07-25
Feature/xivy 14263 improve overwrite dialog

weissreto:
This addresses most of the comments of this PR https://github.com/axonivy/config-editor-client/pull/142:
1. Add more context to the query key
2. Remove not needed methods ' overwritables'
3. Get the knownVariables if the overwrite dialog is opened instead of when the variable editor is loaded
4. Make the overwrite dialog larger

Will try to import more than one variable if one selects a parent in another PR.

### DIFF
--- a/packages/variable-editor/src/VariableEditor.tsx
+++ b/packages/variable-editor/src/VariableEditor.tsx
@@ -31,8 +31,7 @@ function VariableEditor(props: EditorProps) {
     return {
       data: () => genQueryKey('data', context),
       saveData: () => genQueryKey('saveData', context),
-      validate: () => genQueryKey('validate', context),
-      knownVariables: () => genQueryKey('meta/knownVariables', context)
+      validate: () => genQueryKey('validate', context)
     };
   }, [context]);
 
@@ -40,11 +39,6 @@ function VariableEditor(props: EditorProps) {
     queryKey: queryKeys.data(),
     queryFn: () => client.data(context),
     structuralSharing: false
-  });
-
-  const { data: knownVariables } = useQuery({
-    queryKey: queryKeys.knownVariables(),
-    queryFn: () => client.meta('meta/knownVariables', context)
   });
 
   useQuery({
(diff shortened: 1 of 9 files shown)


ivy-lgi (approved):
LGTM :+1:

packages/variable-editor/src/components/variables/master/OverwriteDialog.tsx
```diff
@@ -78,7 +59,31 @@ export const OverwriteDialog = ({ knownVariables, table, variables, setVariables
     }
   };
 
-  const VariableBrowser = ({ applyFn }: { applyFn: (node?: ProjectVarNode) => void }) => {
+  const VariableBrowser = ({ applyFn, context }: { applyFn: (node?: ProjectVarNode) => void; context: DataContext }) => {
+    const client = useClient();
+
+    const queryKeys = useMemo(() => {
+      return {
```
ivy-lli, line 66: if you do nothing else than return something, you can also use the sorthand syntax for it:

```tsx
const queryKeys = useMemo(() => ({ knownVariables: () => genQueryKey('meta/knownVariables', context) }), [context]);
```

packages/variable-editor/src/components/variables/master/OverwriteDialog.tsx
```diff
@@ -78,7 +59,31 @@ export const OverwriteDialog = ({ knownVariables, table, variables, setVariables
...
+
+    const { data: knownVars } = useQuery({
+      queryKey: queryKeys.knownVariables(),
+      queryFn: () => client.meta('meta/knownVariables', context)
+    });
+
+    const [knownVariables, setKnownVariables] = useState<ProjectVarNode>();
+    useEffect(() => {
+      setKnownVariables(knownVars);
+    }, [knownVars]);
```
ivy-lli, line 79: This is not needed. You can directly use the knownVars in the useeffect below

packages/variable-editor/src/VariableEditor.tsx
```diff
@@ -99,11 +94,11 @@ function VariableEditor(props: EditorProps) {
       masterTitle={title}
       masterContent={
         <VariablesMaster
+          context={context}
```
ivy-lli, line 97: instead of move this context prop down, we could also hold this inside of a react context.
But totally fine for now :)

ivy-lli (approved):
Thanks for the improvements, some small final comments.

packages/variable-editor/src/VariableEditor.tsx
weissreto: Yes, I have seen this in inscription view. I will let it like this for now. We can improve if we add more features.

PR #148, opened 2024-07-26
XIVY-14263 Add whole sub tree if one is selected

weissreto:


packages/variable-editor/src/components/variables/master/OverwriteDialog.tsx
```diff
@@ -154,3 +137,30 @@ const toNodes = (root?: ProjectVarNode): Array<BrowserNode> => {
...
+    }
+    return VariableFactory(name);
+  });
+  let childReturnValue = returnValue;
+  for (const child of node.children) {
+    childReturnValue = addVariable(childReturnValue.newData, child);
+  }
+  returnValue.newData = childReturnValue.newData;
+  return returnValue;
+};
```
ivy-lgi, line 166: I think you can remove all the `?` question marks from `node?` since node will always be defined in here.

ivy-lgi (approved):
nice!

Maybe add a test for this new behaviour?
Otherwise good :+1:

ivy-lgi:
> nice!
> 
> Maybe add a test for this new behaviour? Otherwise good 👍

If you move the `addVariable` function to some new file like `project-var-utils.ts` a unit test could be written for this instead of expanding the integration test.

PR #158, opened 2024-09-02
Improve tests

ivy-lgi:


ivy-lgi:
First commit created a lot of changes because of indentation. You can check the second one separately.

ivy-lli (approved):
Nice 👍 
To be clear, it would also be fine for me if you made these adjustments when you edit some tests.

PR #171, opened 2024-09-20
XIVY-15025 Fix input focus to end jump

ivy-lli:
Before:
Screen Recording 2024-09-20 at 08 30 27

After:
Screen Recording 2024-09-20 at 09 46 22

You may ask yourself WTF 🤣 
I was also surprised at the beginning, and I needed some time to find the problem.
The reason is that there are two ways you can write an input field, controlled and uncontrolled:
- controlled: you hold the state -> use the `value` prop
- uncontrolled: the browser holds the state -> use the `defaultValue` prop

So before we used the `value` prop, means we need to hold the state and you may now think we do. I have two assumptions why this is still a problem:
- The reference of the variable changes, so react means there is a new value, and "rerenders to much" of the input so the cursor state forgets where it was
- The parent rerenders and somehow the input losses because of this to much info

And the solution currently for this is:
- Just use the `defaultValue` and let the browser handle the state
- Or hold the state for the input with a useState

As we do not really need to make changes in the UI based on this value state, I made the (in my eyes) the simpler change and just use the `defaultValue` prop, maybe with the "consequence" that we need to change this again in the future, but then also with holding the state for the input field.

PS: by the way this is currently also broken in the dataclass editor, I will provide a fix there too.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -71,17 +71,17 @@
       "link": true
     },
     "node_modules/@axonivy/jsonrpc": {
-      "version": "12.0.0-next.308",
-      "resolved": "https://npmjs-registry.ivyteam.ch/@axonivy/jsonrpc/-/jsonrpc-12.0.0-next.308.tgz",
-      "integrity": "sha512-Jh1akkMx/2E1CaCBfPE6mq0vF6Te1fSYtYPiq6sErBkDP8OM/NM+h7qJkVUpjxkHPzkGsWKdubWVJ46K6G9PmA==",
+      "version": "12.0.0-next.318",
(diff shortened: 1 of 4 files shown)


ivy-lli:
Just realized that the `defaultValue` works but would need a key on the `<VariableDetailsContent key={selectedVariable.join('.')} />` (e.g. key = 0.1.1). 
But there is a problem when we delete (or I assume also add a variable) then the newly selected one has the same key ID as the old one, so the values will still show the old values.
Also set the key e.g. on the full variable path (e.g key = Variables.microsoft-connector.secretKey ) is no solution, because if you change the name then also the full thing gets reloaded and you will loose the focus on the name input. Currently also a problem in the data-class editor I think ;)

So I think I will provide a BasicInput from the ui-components where we hold the state :)

PR #172, opened 2024-10-02
deselect if clicked anywhere in master content

ivy-lgi:
Regarding https://github.com/axonivy/dataclass-editor-client/pull/38
I think I found a better approach while doing the same for the variables editor.
We have an element that triggers the desired `onClick` function and takes the whole space of the master content (`master-content-container`). On the first and only child (`master-content`), we stop the propagation of the event. This way, we don't have to stop it on the rows the control buttons and the search field individually.

### DIFF
--- a/packages/variable-editor/src/components/variables/master/VariablesMasterContent.css
+++ b/packages/variable-editor/src/components/variables/master/VariablesMasterContent.css
@@ -2,7 +2,12 @@
   overflow: hidden auto;
 }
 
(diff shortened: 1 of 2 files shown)


ivy-lli (approved):

ivy-lli:
I think you could also check in the resetSelection handler what the target of the click was, e.g. if it was made to a row do nothing otherwise reset the selection. 
But I like this solution too 👍

PR #185, opened 2024-11-04
Protocol

ivy-rew:
introducing '@axonivy/variable-editor-protocol' ... using the auto-generated types in folllow up PR


ivy-lli (approved):

ivy-lgi (approved):
thanks :+1:

ivy-rew:
> I'm not sure if you really need the vite library build in the protocol package (and the tsconfig.production.json). As technically this package should only contain types right?

Yes, only types. We needed this the last time as work-around when consuming the protocol package in NEO. 
But we can try it without and add it again if really required.

PR #200, opened 2024-11-18
XIVY-15332 Empty value for file meta

ivy-lli:
- Refactor Dialogs

image


packages/variable-editor/src/components/variables/detail/Value.tsx
```diff
@@ -34,6 +34,8 @@ export const Value = ({ variable, onChange }: ValueProps) => {
             onValueChange={(value: string) => onChange([{ key: 'value', value: value }])}
           />
         );
+      case 'file':
+        return <Message message='Value is served by a file.' variant='info' />;
```
ivy-lgi, line 38: I would prefer if there was still an empty, disabled input field.
I know it would serve absolutely no functional purpose, but I think the UI would look better and be more consistent.

ivy-lgi (approved):
One suggestion, otherwise good :+1:

packages/variable-editor/src/components/variables/detail/Value.tsx
ivy-lli: Fine for me 👍

PR #212, opened 2024-12-17
XIVY-15648 meta data details of known variables

ivy-lgi:
The endpoint `knownVariables` has been adapted in the core to return the meta data details of variables of type enum (values) and file (extension). These values are now set on the overwriting variable when using the overwrite dialog.

Additionally:
- Added tests for `nodeIcon`.
- Added tests that editing meta data details is disabled on overwriting variables.
- Minor refactors here and there.

### DIFF
--- a/integrations/standalone/src/mock/meta.ts
+++ b/integrations/standalone/src/mock/meta.ts
@@ -1,43 +1,68 @@
-import type { ProjectVarNode } from '@axonivy/variable-editor-protocol';
+import type { KnownVariables, MetaData } from '@axonivy/variable-editor-protocol';
 
-export const knownVariables: ProjectVarNode = {
-  key: '',
+export const knownVariables: KnownVariables = {
+  namespace: '',
   name: '',
   value: '',
-  type: 'folder',
+  metaData: { type: 'folder' },
   description: '',
   children: [
     {
-      key: 'Amazon',
+      namespace: '',
       name: 'Amazon',
       value: '',
-      type: 'folder',
+      metaData: { type: 'folder' },
       description: '',
       children: [
         {
-          key: 'Amazon.Comprehend',
+          namespace: 'Amazon',
           name: 'Comprehend',
           value: '',
-          type: 'folder',
+          metaData: { type: 'folder' },
           description: 'Amazon comprehend connector settings',
           children: [
             {
-              key: 'Amazon.Comprehend.SecretKey',
+              namespace: 'Amazon.Comprehend',
               name: 'SecretKey',
               value: '<YOUR_SECRET_KEY>',
-              type: 'password',
+              metaData: { type: 'password' },
               description: 'Secret key to access amazon comprehend',
               children: []
             },
             {
-              key: 'Amazon.Comprehend.AccessKey',
+              namespace: 'Amazon.Comprehend',
               name: 'AccessKey',
               value: '<YOUR_ACCESS_KEY>',
-              type: 'string',
+              metaData: { type: 'string' },
               description: 'Access key to access amazon comprehend',
               children: []
             }
           ]
+        }
+      ]
+    },
+    {
+      namespace: '',
+      name: 'Meta',
+      value: '',
+      metaData: { type: 'folder' },
+      description: '',
+      children: [
+        {
+          namespace: 'Meta',
+          name: 'Enum',
+          value: 'two',
+          metaData: { type: 'enum', values: ['one', 'two', 'three'] } as MetaData,
+          description: '',
+          children: []
+        },
+        {
+          namespace: 'Meta',
+          name: 'File',
+          value: '',
+          metaData: { type: 'file', extension: 'json' } as MetaData,
+          description: '',
+          children: []
         }
       ]
     }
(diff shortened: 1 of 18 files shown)


ivy-lgi:
Relevant core PR: https://github.com/axonivy/core/pull/7399

packages/variable-editor/src/components/variables/data/metadata.test.ts
```diff
@@ -54,6 +57,44 @@ describe('isMetadataType', () => {
   });
 });
 
+describe('isMetadata', () => {
+  describe('true', () => {
+    test('empty', () => {
```
ivy-lli, line 62: I know testing ideas or flavours are different.
When I would have written this test, I would have created one test method with every thing in it, e.g. like:


```suggestion
test('isMetadata', () => {
  expect(isMetadata({ type: '' })).toBeTruthy();
  expect(isMetadata({ type: 'password' })).toBeTruthy();
  expect(isMetadata({ type: 'daytime' })).toBeTruthy();
  ...
  expect(isMetadata({ type: 'other' as MetadataType })).toBeFalsy();
  ...
});
```

packages/variable-editor/src/components/variables/dialog/OverwriteDialog.tsx
```diff
@@ -51,29 +52,27 @@ export const OverwriteDialog = ({ table }: OverwriteProps) => {
...
 
-const addVariable = (variables: Array<Variable>, node: ProjectVarNode): AddNodeReturnType<Variable> => {
-  const lastDot = node.key.lastIndexOf('.');
-  const namespace = node.key.substring(0, lastDot);
-  let metadataType: MetadataType = '';
-  if (isMetadataType(node.type)) {
-    metadataType = node.type;
+const addVariable = (variables: Array<Variable>, node: KnownVariables): AddNodeReturnType<Variable> => {
+  let metadata: Metadata = { type: '' };
+  const nodeMetaData = node.metaData as Metadata;
```
ivy-lli, line 57: why is this cast needed?

packages/variable-editor/src/components/variables/dialog/OverwriteDialog.tsx
```diff
@@ -51,29 +52,27 @@ export const OverwriteDialog = ({ table }: OverwriteProps) => {
...
-  const returnValue = addNode(node.name, namespace, variables, name => {
+  let returnValue = addNode(node.name, node.namespace, variables, name => {
     if (name === node.name) {
       return {
         name,
         value: node.value,
         children: [],
         description: node.description,
-        metadata: { type: metadataType }
+        metadata: metadata
```
ivy-lli, line 68: use shorthand

```suggestion
        metadata
```

weissreto (approved):
As far as I understand 🔎👍

integrations/standalone/src/mock/meta.ts
```diff
@@ -1,45 +1,70 @@
...
+      name: 'Meta',
+      value: '',
+      metaData: { type: 'folder' },
+      description: '',
+      children: [
+        {
+          namespace: 'Meta',
+          name: 'Enum',
+          value: 'two',
+          metaData: { type: 'enum', values: ['one', 'two', 'three'] } as MetaData,
```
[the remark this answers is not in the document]
ivy-lgi: In the core, we have a base class `MetaData` just containing the field `type` for meta data types without additional info (`password`, `daytime`) and specific sub classes for meta data types with additional info. E.g. `EnumMetaData` extending `MetaData` and having the additional field `values`.

I assume because the return type of the endpoint only contains the super class `MetaData` but not its sub classes, these classes are not being generated for the protocol type.

I will look into an improvement.

packages/variable-editor/src/components/variables/data/metadata.test.ts
ivy-lgi: I like it better the current way. If something fails, you will immediately know which one did. Are there other advantages/disadvantages?

packages/variable-editor/src/components/variables/dialog/OverwriteDialog.tsx
ivy-lgi: Because the function `isMetadata` (next line) takes an argument of type `Metadata`. Thinking about it, this seems indeed strange. Do you think it's better to change `isMetadata` to take an argument of type `unknown` instead? Something like:
```typescript
export const isMetadata = (metadata: unknown): metadata is Metadata => {
  return (
    typeof metadata === 'object' &&
    metadata !== null &&
    'type' in metadata &&
    isMetadataType((metadata as Metadata).type)
  );
};
```

ivy-lli: Yes definilty. 
The point is that isMetadata is a typeguard right? So it checks if the thing which is given is respecting the wished type. 
So the argument for it should be of type unknown and the return type is guarded as (in this case) Metadata

packages/variable-editor/src/components/variables/data/metadata.test.ts
ivy-lli: if the fail message is good enought I have all the info which I need ;)
Less code and IMO a better overview. 
But as I said this my opinion. I would not say that your way is wrong :)

ivy-lli (approved):

PR #213, opened 2024-12-18
XIVY-15648 import when manually adding variable

ivy-lgi:
When manually adding a variable that is present in a required project, the known variable is imported instead.

import-manually-adding


ivy-lli:
Just from seeing the Gif:
I think this is not a bad idea, however if it works like this, it is not good IMO!
The user just wanted to add a single variable and gets a whole stack? No user will understand what happened.

If we do something like that, at least there should be a warning in the add variable dialog that this variable is known and it will import it!

ivy-lli (approved):
I think the code changes are good, but as said in the comment before, for me something is missing that explains this behavior to the user.

PR #214, opened 2024-12-18
XIVY-15648 info when manually importing variable

ivy-lgi:
An info is shown when trying to manually add a variable that already exists in a required project that it can be imported. Pressing the button again will import the variable.

Also removed some unnecessary pageobject functions.

overwrite-info

### DIFF
--- a/packages/variable-editor/src/components/variables/dialog/AddDialog.tsx
+++ b/packages/variable-editor/src/components/variables/dialog/AddDialog.tsx
@@ -11,16 +11,18 @@ import {
   DialogTrigger,
   Flex,
   Input,
+  Message,
   selectRow
 } from '@axonivy/ui-components';
 import { IvyIcons } from '@axonivy/ui-icons';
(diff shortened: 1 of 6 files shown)


packages/variable-editor/src/components/variables/dialog/AddDialog.tsx
```diff
@@ -106,18 +130,24 @@ export const AddVariableDialog = ({ table }: AddVariableDialogProps) => {
...
-              disabled={!allInputsValid()}
-              onClick={addVariable}
-            >
-              Create Variable
-            </Button>
-          </DialogClose>
+          <Flex>
+            {knownVariable && (
+              <Message variant='info' message='This Variable is already present in a required project. Do you want to import it?' />
+            )}
```
ivy-lli, line 136: I don't like this message left, button right style.
Why not add this message as last child of the main dialog content below the last field.
And I think this can be a warning, then it is more present.

ivy-lgi: image

ivy-lli (approved):

PR #215, opened 2024-12-18
XIVY-15691 do not import variable multiple times

ivy-lgi:


ivy-lli (approved):
Did you check that it is also no possible via the import dialog?

ivy-lgi:
> Did you check that it is also no possible via the import dialog?

The import dialog was the only place where it was possible in the first place.

PR #221, opened 2025-01-06
Update to eslint v9

ivy-lli:


packages/variable-editor/src/utils/tree/tree-data.test.ts
```diff
@@ -177,7 +178,7 @@ describe('removeNode', () => {
       const originalData = structuredClone(data);
       const newData = removeNode(data, [1, 1, 0]);
       expect(data).toEqual(originalData);
-      expect(newData).not.toBe(data);
+      expect(newData).not.toEqual(data);
```
ivy-lgi, line 181: Why make this change in this specific test?

Some info about this line:
These functions under test are designed to not change the data provided as an argument directly. Instead, they create a clone, apply the changes to the clone, and then return the clone. This is important so an object held in state can be provided to the function without having to clone it beforehand.

This line asserts that the returned object of the function is a different one (by reference) than the object that was provided.

Although I think this line is redundant in many cases, I added it to every test just to be safe.

This cloning inside the function strategy is the reason for this whole pattern in every test:
```typescript
const originalData = structuredClone(data); // make a copy of the original data for later use
const newData = removeNode(data, [1, 1, 0]); // call the function under test with the original data
expect(data).toEqual(originalData); // assert that the original data has not been changed
expect(newData).not.toEqual(data); // assert that the returned data is not equal by reference to the original data
```

ivy-lgi (approved):

packages/variable-editor/src/utils/tree/tree-data.test.ts
ivy-lgi: It's fine with me :+1:

PR #222, opened 2025-01-07
Add Navigation via ArrowKey in Main-Table

ivy-edp:
I have added enhanced keyboard support for the main table in the Variable Editor. You can now perform the following actions:
- Navigate through rows using the arrow up/down keys.
- Expand or collapse child rows using the arrow right/left keys.
- Open the inscription view for a selected row by pressing Enter.

keyBoardSupportVariableEditor

### DIFF
--- a/packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
+++ b/packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
@@ -11,12 +11,13 @@ import {
   TableResizableHeader,
   useReadonly,
   useTableExpand,
+  useTableKeyHandler,
   useTableSelect
 } from '@axonivy/ui-components';
 import { IvyIcons } from '@axonivy/ui-icons';
 import { getCoreRowModel, useReactTable, type ColumnDef } from '@tanstack/react-table';
 import { useAppContext } from '../../../context/AppContext';
-import { deleteFirstSelectedRow, useTreeGlobalFilter } from '../../../utils/tree/tree';
+import { deleteFirstSelectedRow, toTreePath, useTreeGlobalFilter } from '../../../utils/tree/tree';
 import { type Variable } from '../data/variable';
 import { variableIcon } from '../data/variable-utils';
 import { AddVariableDialog } from '../dialog/AddDialog';
@@ -25,9 +26,15 @@ import { ValidationRow } from './ValidationRow';
 import './VariablesMasterContent.css';
 
 export const VariablesMasterContent = () => {
-  const { variables, setVariables, setSelectedVariable } = useAppContext();
+  const { variables, setVariables, setSelectedVariable, detail, setDetail } = useAppContext();
 
-  const selection = useTableSelect<Variable>();
+  const selection = useTableSelect<Variable>({
+    onSelect: selectedRows => {
+      const selectedRowId = Object.keys(selectedRows).find(key => selectedRows[key]);
+      const selectedVariable = table.getRowModel().flatRows.find(row => row.id === selectedRowId)?.id;
+      setSelectedVariable(toTreePath(selectedVariable));
+    }
+  });
   const expanded = useTableExpand<Variable>();
   const globalFilter = useTreeGlobalFilter(variables);
   const columns: Array<ColumnDef<Variable, string>> = [
@@ -57,16 +64,19 @@ export const VariablesMasterContent = () => {
     }
   });
 
+  const { handleKeyDown } = useTableKeyHandler({
+    table,
+    data: variables
+  });
+
   const deleteVariable = () =>
     setVariables(old => {
       const deleteFirstSelectedRowReturnValue = deleteFirstSelectedRow(table, old);
-      setSelectedVariable(deleteFirstSelectedRowReturnValue.selectedPath);
       return deleteFirstSelectedRowReturnValue.newData;
     });
 
   const resetSelection = () => {
     selectRow(table);
-    setSelectedVariable([]);
   };
 
   const readonly = useReadonly();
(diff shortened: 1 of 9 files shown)


ivy-edp:
Do you think I should add a playwright test (similar to the Data Class Editor)?

packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
```diff
@@ -56,6 +69,11 @@ export const VariablesMasterContent = () => {
       ...globalFilter.tableState
     }
   });
+  useUpdateSelection(table);
```
ivy-lgi, line 72: I think this might make it obsolete to call `setSelectedVariable` pretty much anywhere else in the project since it's automatically adjusted to the data of the table, right?

Until now, everywhere the data of the table is updated, `setSelectedVariables` is also called explicitly. I think this is no longer necessary with this. You could try removing them.

For example in this file in the methods `deleteVariable` and `resetSelection` but also in other files like the method `updateSelection` in `AddDialog.tsx`, etc.

ivy-lgi (approved):
Some suggestions, otherwise good :+1:

packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
```diff
@@ -23,9 +24,21 @@ import { AddVariableDialog } from '../dialog/AddDialog';
...
+
+const useUpdateSelection = (table: TanstackTable<Variable>) => {
+  const { setSelectedVariable } = useAppContext();
+  const selectedRows = table.getSelectedRowModel().flatRows;
+  const selectedVariable = selectedRows.length === 1 ? selectedRows[0].id : undefined;
+  useEffect(() => {
+    if (selectedVariable) {
+      setSelectedVariable(selectedVariable.split('.').map(Number));
+    }
+  }, [selectedRows, selectedVariable, setSelectedVariable]);
```
ivy-lli, line 37: I just checked that you did this the same way as in the dataclass editor but didn't get it on that review.

I must say that I'm not a fan of this style of selecting things. 
- They lead very very very easy to endless loops which are hard to debug and fix
- not very performant
- more magic than easy code to read

This selection now is somewhere instead where the real action happens. I would be much easier if we could add another callback function to the keyhandler hook (e.g. onSelect) where we can define what should happen if a new selection is triggered.
So if it is not to much effort please consider to improve it, also on the dataclass editor.

packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
ivy-edp: I have also removed setSelectedVariable everywhere else, and it still seems to work well 👍

packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
ivy-edp: I have now extended the useTableSelect hook to allow passing a callback function, which is always executed during onRowSelectChange, eliminating the need for the useEffect hook.

ivy-edp:
configSelectNew

Work well with the new callback-function: The only thing I noticed that seems odd is that when no value is defined, part of the function is directly applied to the value input field of the inscription. This prevents navigating further through the table with the arrow down/up keys. However, I think this can be addressed in a later PR.

packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
```diff
@@ -25,9 +26,15 @@ import { ValidationRow } from './ValidationRow';
...
 export const VariablesMasterContent = () => {
-  const { variables, setVariables, setSelectedVariable } = useAppContext();
+  const { variables, setVariables, setSelectedVariable, detail, setDetail } = useAppContext();
 
-  const selection = useTableSelect<Variable>();
+  const selection = useTableSelect<Variable>({
+    onSelect: selectedRows => {
+      const selectedRowId = Object.keys(selectedRows).find(key => selectedRows[key]);
+      const selectedVariable = table.getRowModel().flatRows.find(row => row.id === selectedRowId)?.id;
+      setSelectedVariable(selectedVariable ? toTreePath(selectedVariable) : []);
```
ivy-lgi, line 35: You could make `toTreePath` accept `undefined` as an argument and then simply change this line to:

```suggestion
      setSelectedVariable(toTreePath(selectedVariable));
```

If `selectedVariable` is `undefined`, `toTreePath` should then return an empty Array (should already be the case).

ivy-lgi (approved):
One small suggestion, otherwise good :+1:

ivy-lgi:
> Work well with the new callback-function: The only thing I noticed that seems odd is that when no value is defined, sometime on row select the focus is set directly on this input field of the inscription. This prevents smoothly navigating further through the table with the arrow down/up keys. However, I think this can be addressed in a later PR.

It's a feature :wink:

See `Value.tsx` (`autoFocus`):
```typescript
          <BasicInput
            value={variable.value}
            onChange={event => onChange([{ key: 'value', value: event.target.value }])}
            autoFocus={variable.value.length === 0}
          />
```
Maybe it's a bad feature now because of arrow navigation, though? :thinking:

packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
ivy-edp: nice, thanks 👍 had to add a undefined in the toTreePath in order to pass a undefined selectedVariable ✌️

ivy-edp:
> > Work well with the new callback-function: The only thing I noticed that seems odd is that when no value is defined, sometime on row select the focus is set directly on this input field of the inscription. This prevents smoothly navigating further through the table with the arrow down/up keys. However, I think this can be addressed in a later PR.
> 
> It's a feature 😉
> 
> See `Value.tsx` (`autoFocus`):
> 
> ```ts
>           <BasicInput
>             value={variable.value}
>             onChange={event => onChange([{ key: 'value', value: event.target.value }])}
>             autoFocus={variable.value.length === 0}
>           />
> ```
> 
> Maybe it's a bad feature now because of arrow navigation, though? 🤔

I was thinking that this is probably a feature. Yes, maybe this needs to be reconsidered now.

PR #226, opened 2025-01-10
XIVY-14401 show validation on input field > master

ivy-lgi:
- Show one validation message with the highest severity on the corrsponding input field.
- Do not show info validation messages in the tree.
- Move methods `validateName` and `validateNamespace` of `validation-utils.ts` to `AddDialog.tsx` since that's the only usage of these methods.
- Make pageobject `Message` part of input fields `TextArea` and `Combobox`.


playwright/tests/pageobjects/Combobox.ts
```diff
@@ -1,20 +1,23 @@
...
     if (options?.label) {
-      this.locator = parentLocator.getByRole('combobox', { name: options.label }).first();
+      this.locator = parent.getByRole('combobox', { name: options.label, exact: true });
     } else {
-      this.locator = parentLocator.getByRole('combobox').nth(options?.nth ?? 0);
+      this.locator = parent.getByRole('combobox').nth(options?.nth ?? 0);
     }
-    this.options = parentLocator.getByRole('option');
-    this.toggleMenu = new Button(parentLocator, { name: 'toggle menu' });
+    this.message = new Message(this.locator.locator('../../..'));
```
ivy-lli, line 18: I think this can easily break...
Better would be to read the `describedby` property and just use it as id for the message.
This also means that you cannot do this here in the constructor but have a message() function that then does this...

playwright/tests/pageobjects/TextArea.ts
```diff
@@ -1,17 +1,20 @@
...
+  constructor(parent: Locator, options?: { label?: string; nth?: number }) {
     if (options?.label) {
-      this.locator = parentLocator.getByLabel(options.label, { exact: true });
+      this.locator = parent.getByLabel(options.label, { exact: true });
     } else {
-      this.locator = parentLocator.getByRole('textbox').nth(options?.nth ?? 0);
+      this.locator = parent.getByRole('textbox').nth(options?.nth ?? 0);
     }
-    this.showPassword = parentLocator.getByLabel('Show password');
+    this.message = new Message(this.locator.locator('..'));
```
ivy-lli, line 16: same here.

ivy-lli (approved):
Cool, just unsure about your usage of XPath locators to get the messages.
I think they can easily break in future, but I must also say that I also have not the perfect other solution...

playwright/tests/pageobjects/Combobox.ts
ivy-lgi: It makes using it a little weird but it's definitely more robust :muscle:

ivy-lli (approved):
Yes I think this is better. Thanks 👍

PR #229, opened 2025-01-15
XIVY-15022 Add shortcuts

ivy-lli:
Global:
- F1: Open help
- A: Add new variable (hold ctrl/cmd to keep it open)
- I: Import variable
- 1-3: jump focus (toolbar, main, inscription)

On Table:
- Delete: remove selected variable

### DIFF
--- a/packages/variable-editor/src/components/variables/dialog/AddDialog.tsx
+++ b/packages/variable-editor/src/components/variables/dialog/AddDialog.tsx
@@ -3,16 +3,20 @@ import {
   Button,
   Combobox,
   Dialog,
-  DialogClose,
   DialogContent,
   DialogFooter,
   DialogHeader,
   DialogTitle,
   DialogTrigger,
   Flex,
+  hotkeyText,
   Input,
   Message,
   selectRow,
+  Tooltip,
+  TooltipContent,
+  TooltipProvider,
+  TooltipTrigger,
   type MessageData
 } from '@axonivy/ui-components';
 import { IvyIcons } from '@axonivy/ui-icons';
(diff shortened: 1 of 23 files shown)


packages/variable-editor/src/components/variables/dialog/AddDialog.tsx
```diff
@@ -137,53 +164,71 @@ export const AddVariableDialog = ({ table }: AddVariableDialogProps) => {
...
   let currentVariables = variables;
   for (const [index, keyPart] of keyParts.entries()) {
+    if (keyPart === '') {
+      return currentVariables;
+    }
     const nextVariable = currentVariables.find(variable => variable.name === keyPart);
     if (nextVariable === undefined) {
-      return;
+      return [];
     }
```
ivy-lgi, line 225: There's a similar method `findVariable` in `known-variables.ts`. Maybe you can move it to somewhere like `variable-utils.ts` and reuse it.

packages/variable-editor/src/components/variables/dialog/AddDialog.tsx
```diff
@@ -137,53 +164,71 @@ export const AddVariableDialog = ({ table }: AddVariableDialogProps) => {
...
     if (!hasChildren(nextVariable)) {
-      return toErrorMessage("Namespace '" + keyParts.slice(0, index + 1).join('.') + "' is not a folder, you cannot add a child to it.");
+      throw new Error("Namespace '" + keyParts.slice(0, index + 1).join('.') + "' is not a folder, you cannot add a child to it.");
     }
```
ivy-lgi, line 228: I would make another method out of everything in this method up to this point and then call that in `validateName` and `validateNamespace` to not have this error-catching logic. Seems weird and hard to read to me.

packages/variable-editor/src/components/variables/dialog/OverwriteDialog.tsx
ivy-lgi, line 1: Importing while holding CTRL to keep the dialog open could also be applied here.

ivy-lgi (approved):
Nice :+1: 

I would make some changes to the `validateName` and `validateNamespace` methods though. Simply have a method `findVariable` and call it in both methods. Then check if it has children in the `validateNamespace` method and if the name is already taken in the `validateName` method.

packages/variable-editor/src/components/variables/dialog/OverwriteDialog.tsx
ivy-lli: Not sure if this has much value...

packages/variable-editor/src/components/variables/dialog/AddDialog.tsx
ivy-lli: yes there is a similar function but simply not the same. findVariable whant one root variable, which I don't have here. 
Also this throw error function (as complicated it is) has a important role, it breaks the finding and fast returns if a namespace is given where one pice is a variable. I'm not sure how I can reuse the same function for both validation methods and still have the same validation result as it is now.

ivy-lli: I tried it quickly but I think it is not that easy to accomplish something the like.

packages/variable-editor/src/components/variables/dialog/AddDialog.tsx
ivy-lli: See comment above. I think it makes sense to have this find logic shared between both validation methods...
And I think it would be sad if we loose this fast error if one variable in the namespace is not a folder but a variable

ivy-lli:
I add a subtask to the story that I check it again if we can simplify the validation code.

Maybe a little bit of context why this change was needed in the first place. I needed to check the validation against the variables, because when you hold the dialog open, the table object is still the old one so no new validation is triggered and you can add multiple times the same variable.
I could even improve the validation a bit with some more cases which where not correct, see the new tests in this area

PR #231, opened 2025-01-16
XIVY-15022 Fix overflow

ivy-lli:
A little bit is cut

Before:
Screenshot 2025-01-16 at 11 03 58

After:
Screenshot 2025-01-16 at 11 03 21


ivy-lgi (commented):
Thanks :+1:

PR #239, opened 2025-01-17
XIVY-15022 Undo / Redo

ivy-lli:
I will move the hook itself and it's tests to the ui-components, as it is more or less the same as in the form editor

And I will also add an undo/redo button I think


ivy-lgi (approved):

ivy-lli:
image

PR #248, opened 2025-01-21
update protocol with polymorphic metadata > 12

ivy-lgi:


ivy-lgi:
Same as https://github.com/axonivy/config-editor-client/pull/247

PR #249, opened 2025-01-21
Update deps

ivy-lli:


ivy-lgi (approved):

ivy-lli:
@ivy-lgi pleas do not blindly merge renovate prs

Especially if there are major jumps (e.g. like react or vite) or if the dependency needs to be the same on all repos (we need to first setup renovate on all, e.g. tanstack query,...)

ivy-lgi:
> @ivy-lgi pleas do not blindly merge renovate prs
> 
> Especially if there are major jumps (e.g. like react or vite) or if the dependency needs to be the same on all repos (we need to first setup renovate on all, e.g. tanstack query,...)

How do I know what I can update and what not? Can you show me?

PR #285, opened 2025-01-23
add renovate range strategy in-range-only

ivy-lgi:


ivy-lgi:
I noticed that when Renovate updates a dependency and the new version is still contained within the range defined in the `package.json`, it only updates the `package-lock.json`. However, it seems to update the range in the `package-lock.json` as well. See for example: https://github.com/axonivy/config-editor-client/commit/e4825124dbe61ed38e487a9249d6c0961e8f9e06.

The next time someone runs `npm i` in the repo, the range in the `package-lock.json` then gets changed again to the range defined in the `package.json`. In the example above, back to `^2.3.1`.

So this value will change multiple times unnecessarily. With this option, Renovate should also update the range in the `package.json`. In the example above to `^2.7.0`.

What do you think?

ivy-lgi:
Actually, this option should be even better as it fixes the issue mentioned above and now only updates within the currently defined range. So, pinned versions will stay pinned, minor ranges will only be updated to other minor ranges, etc.

ivy-lli:
Yes it is normal that bots only update the package.lock. As this on is there real truth and the package.json is more a thought for a range or a minimum point.

Strange that renovate also updates the "copy" of the package.json in the package.lock... but I'm fine if it also updates the package.json

ivy-lgi:
Only problem is that we will no longer be notified about any updates outside of the existing range.

ivy-lli:
> Only problem is that we will no longer be notified about any updates outside of the existing range.

Also not in the dependency track list or in the dependency issue from renovate?

PR #287, opened 2025-01-24
remove horizontal scrollbar in pro designer > master

ivy-lgi:
When a validation message is present, a border is drawn around the respective table row.
In the Pro Designer, this makes a horizontal scrollbar appear on the table, because the border makes the table no longer fit in its container.
This does not happen in Neo or when running the Editor standalone.
I don't understand why this discrepancy exists, but unsetting `overflow-x` on the table makes the scrollbar no longer appear.
However, the border on the right side will not be visible as it does not fit in the container as mentioned above.
No visual change is made to the Neo integration or the standalone Editor.

### DIFF
--- a/packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
+++ b/packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
@@ -17,7 +17,9 @@ import {
 } from '@axonivy/ui-components';
 import { IvyIcons } from '@axonivy/ui-icons';
 import { getCoreRowModel, useReactTable, type ColumnDef } from '@tanstack/react-table';
+import { useRef } from 'react';
 import { useAppContext } from '../../../context/AppContext';
+import { useKnownHotkeys } from '../../../utils/hotkeys';
 import { deleteFirstSelectedRow, toTreePath, useTreeGlobalFilter } from '../../../utils/tree/tree';
 import { type Variable } from '../data/variable';
 import { variableIcon } from '../data/variable-utils';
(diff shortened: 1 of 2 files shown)


ivy-lgi:
Horizontal scrollbar in Pro Designer:
image

I tried some other stuff like `box-sizing: border-box` and playing around with other overflows and the dimensions of the table container, table, rows, etc. but didn't find a way to make the border fit nicely. If you have any other ideas, let me know.

packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
```diff
@@ -114,7 +114,7 @@ export const VariablesMasterContent = () => {
         onClick={event => event.stopPropagation()}
       >
         {globalFilter.filter}
-        <Table onKeyDown={e => handleKeyDown(e, () => setDetail(!detail))}>
+        <Table onKeyDown={e => handleKeyDown(e, () => setDetail(!detail))} style={{ overflowX: 'unset' }}>
```
ivy-lli, line 117: hidden does not work?

ivy-lli (approved):
I think that problem is an inconsistance between browsers. 
Your internal browser is webkit based and I assume you open neo in firefox or chrome.
There is a standalone browser on linux which is also webkit based https://webkit.org/downloads/ the Epiphany. Maybe you can reproduce such issues with this one if you don't want to have this build cycle.

Yes totally fine to disable the horizontal scrolling via overflowX. I assume the problem is an added border or something like that.

packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
ivy-lgi: No, other values that seem to work are:
- clip
- initial
- revert
- visible

ivy-lgi:
> Your internal browser is webkit based and I assume you open neo in firefox or chrome.
> There is a standalone browser on linux which is also webkit based https://webkit.org/downloads/ the Epiphany. Maybe you can reproduce such issues with this one if you don't want to have this build cycle.

You are correct. Using Epiphany I see the same behaviour in Neo and standalone.

PR #288, opened 2025-01-24
remove horizontal scrollbar in pro designer > 12

ivy-lgi:
When a validation message is present, a border is drawn around the respective table row.
In the Pro Designer, this makes a horizontal scrollbar appear on the table, because the border makes the table no longer fit in its container.
This does not happen in Neo or when running the Editor standalone.
I don't understand why this discrepancy exists, but unsetting `overflow-x` on the table makes the scrollbar no longer appear.
However, the border on the right side will not be visible as it does not fit in the container as mentioned above.
No visual change is made to the Neo integration or the standalone Editor.


ivy-lgi:
Same as https://github.com/axonivy/config-editor-client/pull/287

PR #304, opened 2025-02-05
Use shared preset for renovate

alexsuter:
Will use shared config:
https://github.com/axonivy/renovate-config/blob/master/npm.json

- Labels are added in shared preset
- Activating updates on all release branches, but prevent major updates on release branches
- Prevent updating axonivy packages is already defined in shared preset


ivy-lgi (approved):
Thanks :+1:

PR #324, opened 2025-02-07
XIVY-14462 add row virtualization

ivy-lgi:
- Virtualize rows to improve performance.
- Hard code row height to `36px`.
- Set specific values for sizes in the column defintion because the new layout requires it.
- Fix right side of border on rows with messages not being visible.

Existing problems:
- Did not manage to make table header sticky.
- Because of the new layout, the columns are not automatically using all the available space.
  This means that the rightmost column will be cut off even though there seems to be more space available.
  And because of our implementation of `TableResizableHeader`, the rightmost row is currently not resizable.


ivy-lgi:
@ivy-edp in case you're interested.

ivy-lgi:
row-virtualization

packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
```diff
@@ -48,13 +51,22 @@ export const VariablesMasterContent = () => {
...
-      cell: cell => <ExpandableCell cell={cell} icon={variableIcon(cell.row.original)} />,
-      minSize: 50
+      cell: cell => (
+        <ExpandableCell cell={cell} icon={variableIcon(cell.row.original)}>
+          <span>{cell.getValue()}</span>
+        </ExpandableCell>
+      ),
+      minSize: 200,
+      size: 500,
+      maxSize: 1000
```
ivy-lli, line 61: I assume resize is still possible, but this needs to be defined that it works?

packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
```diff
@@ -71,6 +83,25 @@ export const VariablesMasterContent = () => {
...
+      nextElement = nextElement.nextElementSibling;
+    }
+    return height;
+  };
+  const virtualizer = useVirtualizer<HTMLDivElement, HTMLTableRowElement>({
+    count: rows.length,
+    estimateSize: () => ROW_HEIGHT,
+    measureElement,
+    getScrollElement: () => tableContainer.current,
+    overscan: 5
```
ivy-lli, line 102: IMO we could also increase this value, so scrolling looks better, but didn't test it myself.

ivy-lli (approved):
Cool 👍

PR #327, opened 2025-02-10
add namespace to detail view > master

ivy-lgi:
Requested by Bruno.

In a big `variables.yaml`, you can easily get lost where the Variable you have selected is located exactly.

PR #328, opened 2025-02-10
add namespace to detail view > 12

ivy-lgi:
Requested by Bruno.

In a big `variables.yaml`, you can easily get lost where the Variable you have selected is located exactly.


ivy-lgi:
Same as https://github.com/axonivy/config-editor-client/pull/327

PR #336, opened 2025-02-11
XIVY-16068 refactor: prefix css classes > master

ivy-lgi:


packages/variable-editor/src/VariablesEditor.css
```diff
@@ -3,10 +3,15 @@
...
 
-.panel-content-container {
+.variables-editor-panel-content {
   height: 100%;
 }
+
+.variables-editor-help-tooltip-content {
+  font-size: 0.75rem;
+  font-weight: initial;
+}
```
ivy-lli, line 17: technically you don't have to prefix every class. just make the CSS selector more specific:
```css
.variable-editor .master-panel {}
.variable-editor .panel-content-container {}

/* or via CSS nesting */
.variable-editor {
  .master-panel {}
  .panel-content-container {}
}
```

ivy-lli (approved):
Not wrong but I think I would prefer to just change the CSS seletors to be more specific.
When you change all this classes, you may also need to fix tests in neo and vscode, but if you just make the styles more specific when they should match you don't have to rename the classes.

packages/variable-editor/src/VariablesEditor.css
ivy-lgi: I prefer it this way because the classes match exactly. That makes it easier to understand and less error-prone. It also does not rely on another class (`variable-editor`) way up the tree that might be far away from the actual implementation.
That approach will also not work in all cases. For example, dialogs and dropdowns of comboboxes will create elements next to the root, not within it. So the class would not match. So to style something like this you have to go back to this approach anyway, which reduces consistency and might be overlooked, resulting in bugs.

PR #337, opened 2025-02-12
XIVY-16068 refactor: prefix css classes > 12

ivy-lgi:


ivy-lgi:
Same as https://github.com/axonivy/config-editor-client/pull/336 and https://github.com/axonivy/config-editor-client/pull/326

PR #338, opened 2025-02-13
XIVY-16068 refactor: add validations to variables

ivy-lgi:


packages/variable-editor/src/VariablesEditor.tsx
```diff
@@ -162,3 +164,27 @@ function VariableEditor(props: EditorProps) {
...
+export const addValidations = (variables: Array<Variable>, validations: ValidationMessages) => {
+  if (validations.length === 0) {
+    return;
+  }
+  const groupedValidations = groupBy(validations, val => val.path);
+  variables.forEach(variable => {
+    const key = variable.name;
+    variable.validations = groupedValidations[key] ?? [];
+    addValidationsRecursive(variable.children, groupedValidations, key);
+  });
```
ivy-lli, line 177: looks douplicated, cant you just call here directly 
```suggestion
  addValidationsRecursive(variables, groupedValidations, '');
```

ivy-lli: maybe you need to add the dot in the call and not in the function

packages/variable-editor/src/components/variables/data/variable.ts
```diff
@@ -1,16 +1,19 @@
+import type { ValidationMessages } from '@axonivy/variable-editor-protocol';
 import type { TreeNode, TreeNodeFactory, TreeNodeUpdate, TreeNodeUpdates } from '../../../utils/tree/types';
 import type { Metadata } from './metadata';
 
 export interface Variable extends TreeNode<Variable> {
   description: string;
   metadata: Metadata;
+  validations: ValidationMessages;
```
ivy-lli, line 8: why not make this optional, so you don't have to add it to all your test cases?
Technically this would also be fine or not? it has validations (then it is an array) or it has nothing

packages/variable-editor/src/VariablesEditor.tsx
```diff
@@ -115,14 +116,15 @@ function VariableEditor(props: EditorProps) {
     detailTitle += ' - ' + variable.name;
   }
 
+  addValidations(data.root.children, validations);
```
ivy-lli, line 119: I would not add this on this level... only on the data table.

packages/variable-editor/src/components/variables/detail/DetailContent.tsx
```diff
@@ -57,8 +56,6 @@ export const VariablesDetailContent = () => {
   const key = nodesToVariable.map(node => (node ? node.name : ''));
   const overwrites = useOverwrites(key);
 
-  const validations = useValidations(selectedVariable);
-
```
ivy-lli, line 61: I would not change this here. 
If you only add it to the table data you can still do it here the old way. This change is only relevant the the data table so...

ivy-lli (commented):
I don't like this approach of mix it together at the top level. 
This is only a problem for the tree table and should only be solved there...

ivy-lgi:
Replaced by https://github.com/axonivy/config-editor-client/pull/347
Much cleaner and works better.

PR #339, opened 2025-02-14
Move @axonivy packages to peer deps

ivy-lli:


ivy-lli:
Hi @ivy-lgi 
After this change we should be able to update @tanstack/react-query deps via renovate, without fear to break neo.

ivy-lgi (approved):
Cool stuff :+1:

PR #347, opened 2025-02-14
--wip--

ivy-lgi:


packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
```diff
@@ -170,3 +164,33 @@ export const VariablesMasterContent = () => {
...
+  return variables;
+};
+
+const addValidations = (variables: Array<Variable>, groupedValidations: Record<string, ValidationMessages>, currentKey: string) => {
+  variables.forEach(variable => {
+    const key = currentKey + '.' + variable.name;
+    variable.validations = groupedValidations[key] ?? [];
+    addValidations(variable.children, groupedValidations, key);
+  });
+};
```
ivy-lgi, line 188: The main advantage of having two functions is that I don't need to check whether `currentKey` is empty when adding a dot (`.`) as the separator. Otherwise, this check would occur in every recursive call of the function. Now with more logic of checking whether there are any validations in the first place and returning a copy. I think having two functions is definitely justified.

PR #348, opened 2025-02-14
XIVY-16068 improve virtual row height calculation > 12

ivy-lgi:
Instantiate the table with a clone of the variables including their validation messages.
Use these validation messages to calculate the height of the virtual rows without the need to use the DOM elements.
Virtualization seems to work a lot better now.


ivy-lgi:
Same as https://github.com/axonivy/config-editor-client/pull/347

PR #378, opened 2025-02-21
XIVY-15096 adapt to core lsp changes

ivy-lgi:
- Create an abstract `ClientJsonRpc` class that can be implemented by the editors.
- Request objects of new editor LSPs can be added in `protocol/src/client/types.ts`.
- Move `QueryProvider` and `QueryClient` to package `protocol` to be usable by all editors.
- Remove type `ValidationMessages` in favor of using `Array<VariablesValidationResult>` directly.


ivy-lgi:
As discussed, I am going to create separate bundles.

PR #435, opened 2025-03-10
Update renovate config

ivy-lli:
- group non major
- update only to packages which are 3 days old
- only on monday

After that we cannot simply block update via close pr but need to add it here in the config

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,4 +1,9 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": ["local>axonivy/renovate-config:npm"]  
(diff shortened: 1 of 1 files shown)


ivy-lgi:
Is there a reason why we don't want to add this directly in the `renovate-config` repo? https://github.com/axonivy/renovate-config/blob/master/npm.json

ivy-lli:
> Is there a reason why we don't want to add this directly in the `renovate-config` repo? https://github.com/axonivy/renovate-config/blob/master/npm.json

Not sure if this would be the best for all repos... I'm not against rollout this to every repo, but I also like to learn a bit first.

PR #448, opened 2025-03-18
XIVY-16265 I18n translations

ivy-lli:


packages/variable-editor/i18next-parser.config.mjs
```diff
@@ -0,0 +1,11 @@
+export default {
+  defaultNamespace: 'variable-editor',
+  defaultValue: '__MISSING_TRANSLATION__',
+  keepRemoved: false,
```
ivy-fhe, line 4: Would also keep them here.

ivy-fhe (approved):
:+1:

packages/variable-editor/src/components/variables/data/metadata.ts
```diff
@@ -1,7 +1,10 @@
+import { useTranslation } from 'react-i18next';
 import { type VariableUpdate } from './variable';
+import { useMemo } from 'react';
 
 export type Metadata = { type: MetadataType };
-export type MetadataType = (typeof metadataOptions)[number]['value'] | '';
+const metadataOptions = ['', 'default', 'password', 'daytime', 'enum', 'file'] as const;
+export type MetadataType = '' | (typeof metadataOptions)[number];
```
ivy-lgi, line 7: The empty string (`''`) is already present in `metadataOptions`.

packages/variable-editor/src/components/variables/detail/Metadata.tsx
ivy-lgi, line 1: nit: rename file to `MetadataPart.tsx`

packages/variable-editor/src/components/variables/detail/Metadata.tsx
```diff
@@ -3,24 +3,27 @@ import {
...
   type FileMetadataFilenameExtension,
-  type MetadataType
+  type MetadataType,
+  type Metadata
 } from '../data/metadata';
 import { type Variable, type VariableUpdates } from '../data/variable';
 import { EnumValues } from './EnumValues';
+import { useTranslation } from 'react-i18next';
 
 type MetadataProps = {
```
ivy-lgi, line 16: nit: rename to `MetadataPartProps`

eslint.config.mjs
```diff
@@ -12,5 +13,22 @@ export default tseslint.config(
...
+  },
+  {
+    ...i18next.configs['flat/recommended'],
+    files: ['packages/variable-editor/src/**/*.{ts,tsx}'],
+    rules: {
+      'i18next/no-literal-string': [
+        'warn',
+        {
+          mode: 'jsx-only',
+          'jsx-attributes': { include: ['label', 'aria-label', 'title', 'name'] }
```
ivy-lgi, line 25: Maybe add `'message'`? Or does this result in too many false positives?

eslint.config.mjs
```diff
@@ -12,5 +13,22 @@ export default tseslint.config(
...
+        }
+      ]
+    }
+  },
+  {
+    name: 'ignore-files',
+    ignores: ['**/i18next-parser.config.mjs']
```
ivy-lgi, line 32: Does it warn for text directly in an element? e.g. `<DialogTitle>Import Variable</DialogTitle>`. If not, is there a way to make it work?

ivy-lgi (approved):
Some minor suggestions/questions, otherwise good :+1: 
I don't see a test for the translation. Is it possible to add a simple case to show that some label is being translated?

eslint.config.mjs
ivy-lli: yes sadly.

eslint.config.mjs
ivy-lli: yes this will be marked with a warning

PR #495, opened 2025-05-09
Cleanup vite configs

ivy-lli:


packages/variable-editor/vite.config.ts
```diff
@@ -1,15 +1,13 @@
...
-import { defineConfig } from 'vite';
+import { defineConfig } from 'vitest/config';
 import react from '@vitejs/plugin-react';
 import dts from 'vite-plugin-dts';
 
 export default defineConfig({
   plugins: [react(), dts({ tsconfigPath: './tsconfig.production.json' })],
-  esbuild: {
-    sourcemap: 'inline'
-  },
```
ivy-lli, line 10: this includes the sourcemap as inline attachement to the bundled output js, which makes it hughe. Shoud not be a problem for the end output, but cleaner imo

PR #498, opened 2025-05-13
exclude test files in lib output

ivy-lgi:
Actually use `tsconfig.production.json` when packaging.

### DIFF
--- a/packages/variable-editor/package.json
+++ b/packages/variable-editor/package.json
@@ -47,7 +47,7 @@
   "scripts": {
     "clean": "rimraf lib *.tsbuildinfo",
     "build": "tsc --build && tscp",
-    "package": "vite build",
+    "package": "npm run clean && vite build && tsc -p tsconfig.production.json && tscp",
     "build:clean": "npm run clean && npm run build",
     "type": "tsc --noEmit",
     "test": "vitest",
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,16 +38,16 @@
         "@tanstack/react-query": "^5.64",
         "@tanstack/react-query-devtools": "^5.64",
         "i18next": "^25.0.0",
-        "i18next-browser-languagedetector": "^8.1.0",
+        "i18next-browser-languagedetector": "^8.0.4",
         "react": "^19.0.0",
         "react-dom": "^19.0.0",
         "react-i18next": "^15.4.1"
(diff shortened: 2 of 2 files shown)


packages/variable-editor/package.json
```diff
@@ -47,7 +47,7 @@
   "scripts": {
     "clean": "rimraf lib *.tsbuildinfo",
     "build": "tsc --build && tscp",
-    "package": "vite build",
+    "package": "npm run clean && vite build && tsc -p tsconfig.production.json && tscp",
```
ivy-lgi, line 50: What exactly does `tscp` do? Is it needed here?

ivy-lgi: I copied this command from the `form-editor`.

ivy-lgi:
Actually, I don't think this accomplishes what I'm trying to do. @ivy-lmu once mentioned that the editors come with all the test files when used as a dependency in `neo`. I am trying to remove them.

ivy-lli:
> Actually, I don't think this accomplishes what I'm trying to do. @ivy-lmu once mentioned that the editors come with all the test files when used as a dependency in `neo`. I am trying to remove them.

Yes in the src directory all the test file are present, and I think this is ok. What the production.json do is to exclude the test file from transpile to js and also that there are not d.ts file are created from it. So they will not end in the lib output, what is in the end the source of truth for the implementation.

packages/variable-editor/package.json
ivy-lli: https://github.com/body-builder/typescript-cp
We use it or used it to copy e.g. css file to the lib dir as tsc (official typescript compiler/transpiler) only copies or better transpiles ts files to the lib dir.
However since we do a vite lib build I think there should no longer be a need for this (at least where we do this vite lib build)

ivy-lli, line 50: I think this is wrong.
The build already cleans at begin, then do a full tsc build. 
For the real build output however is the dts plugin in the vite.config reliable. 

If there are test files ending in the lib output then it is more likely that some references (e.g. the one in the root tsconfig or the one in the integration) do not refer the production tsconfig and so not filter this files out.

ivy-lgi:
Alright, I learned some stuff. I'm going to look at it on another cool stuff day.

PR #575, opened 2025-08-05
XIVY-17291 Add Importing Variables to the Start View

ivy-tgr:


ivy-tgr:
I couldn't find any test that test this startup view. So should I make one to test the import and also add, or is it not necessary because both features are already tested ?

packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
```diff
@@ -43,6 +43,7 @@ export const VariablesMasterContent = () => {
   const { t } = useTranslation();
   const { variables: originalVariables, setVariables, setSelectedVariable, detail, setDetail, validations } = useAppContext();
   const variables = useMemo(() => variablesWithValidations(originalVariables, validations), [originalVariables, validations]);
+  const { importVar: shortcut } = useKnownHotkeys();
```
ivy-lli, line 46: not needed, see other comment

packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
```diff
@@ -122,7 +123,9 @@ export const VariablesMasterContent = () => {
         <Button className='variables-editor-add-button' icon={IvyIcons.Plus} aria-label={hotkeys.addVar.label} />
       </AddVariableDialog>
       <Separator decorative orientation='vertical' style={{ height: '20px', margin: 0 }} />
-      <OverwriteDialog table={table} />
+      <OverwriteDialog table={table}>
+        <Button icon={IvyIcons.FileImport} aria-label={shortcut.label} />
```
ivy-lli, line 127:
```suggestion
        <Button icon={IvyIcons.FileImport} aria-label={hotkeys.importVar.label} />
```

ivy-lli (approved):
Perfect! Once small improvement suggestion below.
And yes it seems to have no test yet, if you like it would be great to have one test which quickly test that on an empty editor those tow dialog can be opened via buttons and via shortcuts 👍

PR #588, opened 2025-08-22
Use shared tsconfig

ivy-lli:


packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
```diff
@@ -48,7 +48,9 @@ export const VariablesMasterContent = () => {
     onSelect: selectedRows => {
       const selectedRowId = Object.keys(selectedRows).find(key => selectedRows[key]);
       const selectedVariable = table.getRowModel().flatRows.find(row => row.id === selectedRowId)?.id;
-      setSelectedVariable(toTreePath(selectedVariable));
+      if (selectedVariable || selectedRowId === undefined) {
+        setSelectedVariable(toTreePath(selectedVariable));
```
ivy-lgi, line 52: Seems like a weird logic sequence.
First, `selectedRowId` is used in another operation and then it is checked whether it's defined. I assume the case `selectedRowId === undefined` could be handled directly after its instantiation, before it is used, or not?

packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
```diff
@@ -93,7 +95,7 @@ export const VariablesMasterContent = () => {
   const tableContainer = useRef<HTMLDivElement>(null);
   const virtualizer = useVirtualizer<HTMLDivElement, HTMLTableRowElement>({
     count: rows.length,
-    estimateSize: index => rowHeight(rows[index].original.validations),
+    estimateSize: index => rowHeight(rows[index]?.original.validations),
```
ivy-lgi, line 98: Is there a reason to not use `!` at all? e.g. `rows[index]!.original.validations`
I think in this statement it should always be defined. Otherwise, something went really wrong.
That might make it clear to other developers that it's always defined.

ivy-lgi (approved):

packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
ivy-lli: hm need to check again.
This fix is for: on the dev (not prod) after an variable is added it is selected in the table but not in the context, and because of this it is not shown in the detail view.

packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
ivy-lli: https://typescript-eslint.io/rules/no-non-null-assertion/

ivy-lli: Either we accept this fact, but I think "trick" typescript is not really the goal here... Especially it leads that just everyone will add a `!`, instead of really check if the array element is not undefined.
Maybe it is sometimes already checked, but in a case where typescript doesn't know it, but mostly those cases can be written a bit different.

packages/variable-editor/src/components/variables/master/VariablesMasterContent.tsx
ivy-lli: the problem here is that this code will be triggered multiple times, the selectedRows is already set, but the table row model has not yet this new row. 
So when the selectedRowId has a value, but the selectedVariable is undefined then do nothing
If the selectedRowId is undefined then the table has cleared all its selections.

but yes I can write it a bit different

